PR #12, opened 2020-08-23
refactor: unhardcoded weth and protocol token addresses and fix a test

kurt-wagner:
unhardcoded weth and protocol token addresses
fix test_marginTradeFromPool_sim

I used `sovryn` as protocol token in `scripts/deploy_everything_local.py` file because I wasn't sure which token should use.

### DIFF
--- a/contracts/core/Constants.sol
+++ b/contracts/core/Constants.sol
@@ -6,15 +6,25 @@
 pragma solidity 0.5.17;
 
 import "../interfaces/IWethERC20.sol";
-
+import "../openzeppelin/Address.sol";
 
 contract Constants {
-    //IWethERC20 public constant wethToken = IWethERC20(0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2); // mainnet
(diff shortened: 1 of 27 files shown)


ororopickpocket:
We do not have an own protocol token, yet. For now it can just be any address.

There's one thing which I don't like. Now, we need to pass the same contract addresses to a lot of contracts on deployment. Can you think of a way to set it only 1x for all contracts (at least the protocol modules) similar to how the other state variables are shared (State.sol)?

kurt-wagner:
> We do not have an own protocol token, yet. For now it can just be any address.
> 
> There's one thing which I don't like. Now, we need to pass the same contract addresses to a lot of contracts on deployment. Can you think of a way to set it only 1x for all contracts (at least the protocol modules) similar to how the other state variables are shared (State.sol)?

Protocol module contracts handle this addresses as a configuration in the ProtocolSettings contract
I left priceFeed as it is. Should those addresses be configurable? Let me know an I change that

ororopickpocket:
Isn't it possible to let VaultController inherit from State instead of Constants and then not pass the weth contract as parameter on the function calls? Is VaultController only called via delegatecall?

PR #29, opened 2020-09-08
Wrbtc integration

ororopickpocket:


contracts/connectors/loantoken/LoanToken.sol
```diff
@@ -10,20 +10,20 @@ import "./AdvancedTokenStorage.sol";
 
 contract LoanToken is AdvancedTokenStorage {
 
-    // It is important to maintain the variables order so the delegate calls can access sovrynContractAddress and wethTokenAddress
+    // It is important to maintain the variables order so the delegate calls can access sovrynContractAddress and wbtcTokenAddress
```
kurt-wagner, line 13: should be `wrbtc` or is `wbtc`  alright?

contracts/swaps/connectors/testnet/SwapsImplLocal.sol
```diff
@@ -25,11 +25,9 @@ contract SwapsImplLocal is State, ISwapsImpl {
         uint256 requiredDestTokenAmount)
         public
         returns (uint256 destTokenAmountReceived, uint256 sourceTokenAmountUsed)
-    {
+    {   
         require(sourceTokenAddress != destTokenAddress, "source == dest");
-        require(sourceTokenAddress != address(wethToken) && destTokenAddress != address(wethToken),
-            "WETH swaps not supported on testnet"
```
kurt-wagner, line 31: is this ok?

kurt-wagner (approved):

contracts/connectors/loantoken/LoanToken.sol
ororopickpocket: it is wrapped rBTC of course, but we try to give the user the impression he's working directly with BTC (although of course, he's working with rBTC). That's why I called it wBTC and not wRBTC. That's also why on the frontend rBTC was replaced with BTC. But it's a good point and worth getting a confirmation from Edan on the hidden core channel.

contracts/swaps/connectors/testnet/SwapsImplLocal.sol
ororopickpocket: yes, because i adapted the TestWBTC (formerly TestWETH) contract to also allow arbitrary minting and burning.

ororopickpocket: needed it for testing

PR #31, opened 2020-09-11
feat: change rollover reward

kurt-wagner:


kurt-wagner:
> Is there a test making sure the correct reward is being paid on rollover?

oh, I forgot that part. I'll add it

PR #32, opened 2020-09-11
feat: add early access token - smart contract adaption

kurt-wagner:
As I added OpenZepelin as a dependency some contracts produce name collision issues. So I change their names: 

- Context.sol
- SafeMath.sol
- IERC20.sol

Shortly, we can use OpenZeppelin contracts directly and remove the old ones. I did not do that in this PR because the changes were not so trivial.

If you get an error like this one:
`FileNotFoundError: [Errno 2] No such file or directory: 'contracts/openzeppelin/SafeMath.sol'`
Run: `rm -rf build  .pytest_cache __pycache__`


ororopickpocket:
out of interest: what were the issues when trying to replace the existing contracts with those from open zeppelin?

kurt-wagner:
> out of interest: what were the issues when trying to replace the existing contracts with those from open zeppelin?

NamespaceCollision: Multiple contracts or interfaces with the same name
  ContextSovryn: contracts/openzeppelin/Context.sol, contracts/openzeppelin/ContextSovryn.sol

ororopickpocket:
not going to be added to the standard code base, but is merged into the mainnet deployment branch

PR #33, opened 2020-09-14
Deposit collateral event

ororopickpocket:


contracts/events/LoanMaintenanceEvents.sol
```diff
@@ -0,0 +1,10 @@
+pragma solidity 0.5.17;
+
+contract LoanMaintenanceEvents {
+    
+    event DepositCollateral(
```
kurt-wagner, line 5: Would not be a good idea to add the sender?

tests/protocol/addingMargin/test_deposit_collareral_using_TestToken.py
```diff
@@ -17,14 +17,20 @@ def test_deposit_collateral(sovryn,set_demand_curve,lend_to_pool,open_margin_tra
...
+    deposit_amount = startCollateral/2
     
     #deposit collateral to add margin to the loan created above
-    RBTC.approve(sovryn, startCollateral/2)
-    sovryn.depositCollateral(loan_id, startCollateral/2)
+    RBTC.approve(sovryn, deposit_amount)
+    tx = sovryn.depositCollateral(loan_id, deposit_amount)
+    
+    #verify the deposit collateral event
+    print(tx.info())
```
kurt-wagner, line 27: You don't need to use `print`. It also works just with `tx.info()`

kurt-wagner (approved):

contracts/events/LoanMaintenanceEvents.sol
ororopickpocket: Victor just asked for loanId and amount, but yes, maybe it'd turnout helpful at some point in the future to have the sender as well.

tests/protocol/addingMargin/test_deposit_collareral_using_TestToken.py
ororopickpocket: thanks. if you've got more hints, go ahead ;)

ororopickpocket:
The toggle function pause is now not working anymore. that's because it's now handled differently on the settings contract, but the logic contract wasn't updated (checkPause is not working anymore). Could you also merge the changes of the loan token logic contract?

PR #34, opened 2020-09-15
feat: merge changes from bzx repo for LoanToken.sol and LoanTokenSett…

kurt-wagner:
…ingsLowerAdmin.sol


ororopickpocket:
The toggle function pause is now not working anymore. that's because it's now handled differently on the settings contract, but the logic contract wasn't updated (checkPause is not working anymore). Could you also merge the changes of the loan token logic contract?

PR #39, opened 2020-09-23
doc: add Readme with tests common issues and how to solve them

kurt-wagner:


ororopickpocket:
Thanks, Kurt, but could you also leave hints for the tests which are sometimes failing, sometimes not?

PR #58, opened 2020-11-12
Add BPro price feed

wukong-xxs:


ororopickpocket (changes requested):
Everything good, just 1 little request: Please move the javascript test file to a new folder tests-js to not mix it up with the python (the governance contract tests are also written in JS and will be in the same folder)

ororopickpocket:
All of the code included in add-usdt

PR #67, opened 2020-11-22
Timelock error message propagation

jameshowlett977:
Error message wasn't propagated, so we had the following message for all failed transactions:
`Timelock::executeTransaction: Transaction execution reverted.`

With error message propagation:
`Timelock::executeTransaction: Timelock::setDelay: Delay must exceed minimum delay.`
`Timelock::executeTransaction: unauthorized`


ororopickpocket (approved):
nice!

PR #71, opened 2020-11-27
Refactor loan token settings handling

wukong-xxs:


contracts/connectors/loantoken/LoanTokenLogicStandard.sol
```diff
@@ -7,16 +7,15 @@ pragma solidity 0.5.17;
...
 
 import "./AdvancedToken.sol";
+import "./LoanTokenSettingsLowerAdmin.sol";
 import "./interfaces/ProtocolLike.sol";
 import "./interfaces/FeedsLike.sol";
 
 
-contract LoanTokenLogicStandard is AdvancedToken {
-    using SafeMath for uint256;
+contract LoanTokenLogicStandard is AdvancedToken, LoanTokenSettingsLowerAdmin {
```
ororopickpocket, line 15: LoanTokenSettingsLowerAdmin is inheriting from AdvancedTokenStorage, AdvancedToken is also inheriting from AdvancedTokenStorage.... I think it's better to let LoanTokenSettingsLowerAdmin inherit from AdvancedToken instead.

contracts/connectors/loantoken/LoanTokenSettingsLowerAdmin.sol
```diff
@@ -133,4 +133,9 @@ contract LoanTokenSettingsLowerAdmin is AdvancedTokenStorage {
         emit SetTransactionLimits(addresses, limits);
     }
 
+    function changeLoanToken(string memory _name, string memory _symbol) public onlyAdmin {
```
ororopickpocket, line 136: Confusing naming. Call it changeLoanTokenName

contracts/connectors/loantoken/LoanTokenLogicStandard.sol
ororopickpocket: Please, also remove the function called 'updateSettings'. It's not needed anymore.

tests/test_modifying_name_and_symbol.py
```diff
@@ -0,0 +1,15 @@
+import pytest
+from brownie import Contract, Wei, reverts
+from fixedint import *
+import shared
+
+def test_modifying_name_and_symbol(accounts, loanToken, loanTokenLogic, name, symbol, LoanToken, LoanTokenLogicStandard):
+    
+    localLoanToken = Contract.from_abi("loanToken", address=loanToken.address, abi=LoanToken.abi, owner=accounts[0])
+    localLoanToken.setTarget(loanTokenLogic.address)
+    localLoanToken = Contract.from_abi("loanToken", address=loanToken.address, abi=LoanTokenLogicStandard.abi, owner=accounts[0])
```
ororopickpocket, line 10: This is not necessary, because we don't have different targets anymore (token logic is the target by default).

Delete the lines  8 + 9

Do the same for all of the other test cases you modified. No more settings targets, just load the loanToken directly with the logic ABI.

ororopickpocket (changes requested):

ororopickpocket:
You forgot to update the loan token deployment script

scripts/deployment/deploy_loanToken.py
```diff
@@ -93,57 +93,55 @@ def deployLoanToken(acct, sovryn, loanTokenAddress, loanTokenSymbol, loanTokenNa
...
+            collateralAddress, ## collateralToken.
+            Wei("20 ether"), ## minInitialMargin -> 20% (allows up to 5x leverage)
+            Wei("15 ether"), ## maintenanceMargin -> 15%, below liquidation
+            0 ## fixedLoanTerm -> will be overwritten with 28 days
+        ]
+
+        params.append(data)
+
+    #configure the token settings, and set the setting contract address at the loan token logic contract
+    tx = loanToken.setupLoanParams(params, False)
```
ororopickpocket, line 120: One question, Wukong. Here, you pass (params, False). Before it was 
```calldata = loanTokenSettings.setupLoanParams.encode_input(params, True)```
What difference does it make?

wukong-xxs: before:
`calldata = loanTokenSettings.setupLoanParams.encode_input(params, False)`
`tx = loanToken.updateSettings(loanTokenSettings.address, calldata, { "from": acct })`

now:
`tx = loanToken.setupLoanParams(params, False)`

nothing different. just directly call setupLoanParams() of loanToken

ororopickpocket: True != False.

ororopickpocket: look at the params

ororopickpocket: sorry, my bad. i was comparing the wrong lines. everything is fine!

PR #72, opened 2020-11-30
[DON'T MERGE] Vesting tests + some fixes

havok1231:


contracts/governance/Vesting/Vesting.sol
```diff
@@ -101,15 +105,19 @@ contract Vesting is Ownable{
                 staking.withdraw(stake, i, receiver);
         }
     }
-    
+
+    /**
+     * @dev collect dividends from fee sharing proxy
+     */
     function collectDividends() public onlyOwners{
```
ororopickpocket, line 112: The fees should be withdrawn to the token owner

ororopickpocket: maybe we can add a receiver to the withdraw function of the fee sharing proxy?

tests-js/vesting/Vesting.js
```diff
@@ -18,114 +18,210 @@ const Vesting = artifacts.require('Vesting');
...
     });
-    
+
     describe('stakeTokens', () => {
+        let vesting;
         it('should stake 1,000,000 SOV with a duration of 104 weeks and a 26 week cliff', async () => {
+            vesting = await Vesting.new(token.address, staking.address, root, new BN(26 * 7 * 24 * 60 * 60), new BN(104 * 7 * 24 * 60 * 60), feeSharingProxy);
 
+            await token.approve(vesting.address, ONE_MILLON);
+            await vesting.stakeTokens(ONE_MILLON);
```
ororopickpocket, line 88: also check that the positions exist on the staking contract and that the voting right is delegated to the token owner

tests-js/vesting/Vesting.js
```diff
@@ -18,114 +18,210 @@ const Vesting = artifacts.require('Vesting');
...
             //1. set new staking contract address on staking contract
+
+            let newStaking = await StakingProxy.new(token.address);
+            await newStaking.setImplementation(stakingLogic.address);
+            newStaking = await StakingLogic.at(newStaking.address);
+
+            await staking.setNewStakingContract(newStaking.address);
+
             //2. call migrateToNewStakingContract
+            await vesting.migrateToNewStakingContract();
```
ororopickpocket, line 200: verify that the new staking contract was set on the vesting contract

ororopickpocket (changes requested):

ororopickpocket:
good and quick work in general!

contracts/governance/Vesting/Vesting.sol
havok1231: Sorry, I can't correctly work with 2 different branches. I could change it in IFeeSharingProxy, but how it'll be implemented I don't know.
I'll add @todo in IFeeSharingProxy for now.

tests-js/vesting/Vesting.js
havok1231: this is not implemented in Staking contract.

function setNewStakingContract(address _newStakingContract) public onlyOwner {
        require(_newStakingContract != address(0), "can't reset the new staking contract to 0");
        newStakingContract = _newStakingContract;
    }

function migrateToNewStakingContract() public {
        require(newStakingContract != address(0), "there is no new staking contract set");
        //implementation:
        //iterate over all possible lock dates from now until now + MAX_DURATION
        //read the stake & delegate of the msg.sender
        //if stake > 0, stake it at the new contract until the lock date with the current delegate
    }

tests-js/vesting/Vesting.js
havok1231: I wanted to do this, but "staking" is a private variable. I can using getStorageSlot, but in test it will looks weird.

havok1231: also all variables in Vesting contract are public, don't looks like you missed "public" for staking. I can add it, if it neccessary.

tests-js/vesting/Vesting.js
havok1231: it should be tested on the Stake contract

tests-js/vesting/Vesting.js
havok1231: made it public. Done.

contracts/governance/Vesting/Vesting.sol
havok1231: Done. Changed to this signature:
function withdrawToken(address _token, uint32 _maxCheckpoints, address _receiver) public;.

tests-js/vesting/Vesting.js
havok1231: Done

jameshowlett977 (approved):

PR #74, opened 2020-12-01
adoption vesting procedure (RSOV)

jameshowlett977:


jameshowlett977:
Added dev dependecy: "openzeppelin-solidity": "^2.5.1"

contracts/governance/Vesting/RSOV.sol
```diff
@@ -0,0 +1,101 @@
...
+        emit Mint(msg.sender, _amount);
+    }
+    
+    /**
+     * @notice burns RSOV tokens and stakes the respective amount SOV tokens in the user's behalf
+     * @param _amount the amount of tokens to be burnt
+     */
+    function burn(uint _amount) public {
+        require(_amount > 0, "RSOV:: burn: amount invalid");
+        uint96 amount96 = safe96(_amount, "RSOV::burn: amount exceeds 96 bits");
```
ororopickpocket, line 78: why not make _amount uint96 in the first place?

ororopickpocket (approved):

contracts/governance/Vesting/RSOV.sol
jameshowlett977: Do you mean make it uint96 before `require`?

contracts/governance/Vesting/RSOV.sol
```diff
@@ -0,0 +1,101 @@
...
+        
+        //burns RSOV tokens
+        _burn(msg.sender, _amount);
+        
+        //stakes SOV tokens in the user's behalf
+        SOV.approve(address(staking), _amount);
+        uint until = block.timestamp + stakingDuration;
+        staking.stake(amount96, until, msg.sender, msg.sender);
```
ororopickpocket, line 86: Forhot about this. here we need to check first if the user already staked until this time and if he did increase the stake instead

jameshowlett977: Changed Staking contract as we agreed.
Left public function `increaseStake` that invokes `_increaseStake`.
I wasn't not sure whether we need to remove it (I think frontend use it).

contracts/governance/Staking/Staking.sol
```diff
@@ -22,29 +22,35 @@ contract Staking is WeightedStaking{
...
+            //change delegate if new one is specified
             _delegate(stakeFor, delegatee, until);
-        
-        emit TokensStaked(stakeFor, amount, until, amount);
+            //increase stake, new delegate will be used
+            _increaseStake(amount, stakeFor, until);
+        } else {
+            //do not stake longer than the max duration
+            if (until > block.timestamp + MAX_DURATION)
+                until = block.timestamp + MAX_DURATION;
```
ororopickpocket, line 37: this needs to be done before the if-else. It could be that the user already has tokens staked until that point. 

additional optional improvement suggestion:
The same code which is in the else is also part of _increaseStake. I think we can avoid code duplication by calling ```_icreaseStake``` in any event and just let it contain the steps whcih are the same for both functions: ```increaseStake``` and ```stake```.

ororopickpocket (changes requested):

contracts/governance/Staking/Staking.sol
jameshowlett977: oh, thanks

refactored

contracts/governance/Staking/Staking.sol
```diff
@@ -20,34 +20,38 @@ contract Staking is WeightedStaking{
...
+        uint96 previousBalance = currentBalance(stakeFor, until);
+        //increase stake
+        _increaseStake(amount, stakeFor, until);
+
+        if (previousBalance == 0) {
+            //regular delegation if it's a first stake
             _delegate(stakeFor, delegatee, until);
-        
-        emit TokensStaked(stakeFor, amount, until, amount);
+        } else {
```
ororopickpocket, line 42: nice, you're even allowing to change the delegate in the same transaction.

contracts/governance/Staking/Staking.sol
```diff
@@ -90,32 +93,78 @@ contract Staking is WeightedStaking{
      * @param amount the amount of SOV tokens
      * @param stakeFor the address for which we want to increase the stake. staking for the sender if 0x0
      * @param until the lock date until which the funds are staked
-     */
-    function increaseStake(uint96 amount, address stakeFor, uint until) public{
+     * */
+    //TODO deprecated, can be removed when frontend updated
```
ororopickpocket, line 97: I think we can remove it immediatley. i'll let victor know that the next staking contract will no longer support this function.

contracts/governance/Staking/Staking.sol
```diff
@@ -90,32 +93,78 @@ contract Staking is WeightedStaking{
...
+    /**
+     * @notice stakes tokens according to the vesting schedule
+     * @param amount the amount of tokens to stake
+     * @param cliff the time interval to the first withdraw
+     * @param duration the staking duration
+     * @param intervalLength the length of each staking interval when cliff passed
+     * @param stakeFor the address to stake the tokens for or 0x0 if staking for oneself
+     * @param delegatee the address of the delegatee or 0x0 if there is none.
+     * */
+    function stakeTokens(
```
ororopickpocket, line 140: maybe we can rename this function, so it's clearer what it does? like stakeBySchedule? or stakeVested?

contracts/governance/Staking/Staking.sol
```diff
@@ -90,32 +93,78 @@ contract Staking is WeightedStaking{
...
+        if(numIntervals > 1) {
+            stake(uint96(amount - stakedPerInterval * (numIntervals-1)), start, stakeFor, delegatee);
+        }
+        //stake the rest in 4 week intervals
+        for(uint i = start + intervalLength; i <= end; i+= intervalLength) {
+            //stakes for itself, delegates to the owner
+            stake(uint96(stakedPerInterval), i, stakeFor, delegatee);
+        }
+    }
+
```
ororopickpocket, line 167: has nothing to do with this PR, but I realized that we're not reducing the delegate stake when withdrawing tokens.

contracts/governance/Staking/Staking.sol
jameshowlett977: thanks

contracts/governance/Staking/Staking.sol
jameshowlett977: done

contracts/governance/Staking/Staking.sol
jameshowlett977: renamed to stakeBySchedule

contracts/governance/Staking/Staking.sol
jameshowlett977: fixed, tests updated

ororopickpocket (approved):

PR #77, opened 2020-12-09
Early unstaking

jameshowlett977:


contracts/governance/FeeSharingProxy.sol
```diff
@@ -69,11 +81,53 @@ contract FeeSharingProxy is SafeMath96, IFeeSharingProxy {
...
+        unprocessedAmount[_token] = add96(unprocessedAmount[_token], _amount, "FeeSharingProxy::transferTokens: amount exceeds 96 bits");
+
+        emit TokensTransferred(msg.sender, _token, _amount);
+    }
+
+    /**
+     * @notice adds checkpoint with accumulated amount by function invocation
+     * @param _token address of the token
+     * */
+    function addCheckpoint(address _token) public {
```
ororopickpocket, line 114: why not internal?

(no full review at this point, just quick scan)

jameshowlett977: - `transferTokens` will be invoked by Staking contract for all early unstaking
- `addCheckpoint` should be invoked externally in order to create checkpoint (currently, we allow to create new checkpoint after 24h after previous one), we need to think about scheduling it
- `withdraw` will be invoked by user, all newly created and unprocessed checkpoint should be processed here

ororopickpocket: why do we need to add a checkpoint externally?
don't we only need it after withdrawing fees and transferring tokens?

ororopickpocket:
```
Compiling your contracts...
===========================
Error: Could not find openzeppelin-solidity/contracts/token/ERC20/ERC20Detailed.sol from any sources; imported from /home/ubuntu/remake/Sovryn-smart-contracts/contracts/governance/Vesting/RSOV.sol
```

contracts/governance/FeeSharingProxy.sol
jameshowlett977: We don't want to create a checkpoint for every slashing of tokens.
I think it's better to invoke `addCheckpoint` from `transferTokens` if 24h passed since last checkpoint.
Can change it tomorrow.

jameshowlett977:
Package `openzeppelin-solidity` was added. Please run `npm install`.
I think about moving a few contracts to `contracts/openzeppelin` or re-configuring brownie (trufflle compiles, brownie doesn't compile).

contracts/governance/FeeSharingProxy.sol
jameshowlett977: done

contracts/governance/FeeSharingProxy.sol
```diff
@@ -67,13 +75,50 @@ contract FeeSharingProxy is SafeMath96, IFeeSharingProxy {
...
         IERC20(_token).approve(loanToken, amount);
         uint poolTokenAmount = ILoanToken(loanPoolToken).mint(address(this), amount);
-        _writeTokenCheckpoint(loanPoolToken, safe96(poolTokenAmount, "FeeSharingProxy::withdrawFees: pool token amount exceeds 96 bits"));
-
-        lastFeeWithdrawalTime = block.timestamp;
-        
+        if (block.timestamp - lastFeeWithdrawalTime[loanPoolToken] >= FEE_WITHDRAWAL_INTERVAL) {
+            lastFeeWithdrawalTime[loanPoolToken] = block.timestamp;
+            _writeTokenCheckpoint(loanPoolToken, safe96(poolTokenAmount, "FeeSharingProxy::withdrawFees: pool token amount exceeds 96 bits"));
+        }
```
ororopickpocket, line 81: this worries me. it makes sense to allow fee withdrawal also between checkpoints, but here we're writing a checkpoint only for the pool tokens received on this transaction. what happens with pool tokens which get received in between checkpoints? I don't see this function remembering the unprocessed amount.

contracts/governance/Staking/Staking.sol
```diff
@@ -302,4 +369,36 @@ contract Staking is WeightedStaking{
         allUnlocked = true;
         emit TokensUnlocked(SOVToken.balanceOf(address(this)));
     }
+
+    /**
+     * @notice Gets list of stakes for `account`
+     * @param account The address to get stakes
+     * @return The arrays of dates and stakes
+     */
+    function getStakes(address account) public view returns (uint[] memory dates, uint96[] memory stakes) {
```
ororopickpocket, line 378: maybe make this external? it's only for frontend usage, isn't it?

contracts/governance/FeeSharingProxy.sol
jameshowlett977: You're right, thanks.
Updated test to catch this issue and fixed contract.

contracts/governance/Staking/Staking.sol
jameshowlett977: yes, done

ororopickpocket (approved):

PR #79, opened 2020-12-09
Use AMM rate

wukong-xxs:


ororopickpocket:
@wukong-xxs on my last commit you can see the solution for the problem with the reverting calls.
The problem was that the ```sovrynSwapContractRegistryAddress``` on ```SwapsImplSovryn``` was expected to be read from the sovryn protocol context. When performing a swap, the sovryn protocol delegatecalls the ```internalSwap``` and then, the address is read from the ```State.sol```. When calling ```internalExpectedRate``` with a simple call, ```sovrynSwapContractRegistryAddress``` was the 0 address. Therefore, the call failed.

ororopickpocket:
It took me nearly the whole day to realize that this task is stupid. 
I was able to solve the issues, but it still didn't make sense. In all of the cases we need the inverse rate.
We want to know how many tokens X do I have to spend to get the right amount of tokens Y, and the swap cannot give us this answer. 
If we query the rate from Y to X from the swap, our estimation will be farther off from the actual result than if using the price feed.

ororopickpocket:
reopened because some of the fixes are needed anyway

ororopickpocket:
New plan: Query the rate with the price feed, then get the expected return for the result. compare to the desired value and scale accordingly.

contracts/connectors/loantoken/LoanTokenLogicStandard.sol
```diff
@@ -705,6 +705,51 @@ contract LoanTokenLogicStandard is LoanTokenSettingsLowerAdmin {
...
+            collateralToken, 
+            loanTokenAmount);
+        
+        // compares both. if different, scales loan size accordingly
+        if (collateralAmount != collateralAmountExpected) {
+            loanTokenAmount = ProtocolLike(sovrynContractAddress).getSwapExpectedReturn(
+                collateralToken, 
+                loanToken, 
+                collateralAmountExpected);
+        }
```
ororopickpocket, line 748: this needs to convert the loan token amount which you will calculate above back to collateral tokens. this should not be in this ```if``` block

contracts/connectors/loantoken/LoanTokenLogicStandard.sol
```diff
@@ -705,6 +705,51 @@ contract LoanTokenLogicStandard is LoanTokenSettingsLowerAdmin {
...
+
+        return loanTokenAmount;
```
ororopickpocket, line 750: scaling is missing completely. 

At this point you've got 2 numbers: the collateral tokens which were sent and the collateral tokens you'd receive if using the loantokenamount you calculated earlier.

if the calculated collateral token amount is bigger than the one passed as param, you need to scale the loanTokenAmount up accordingly, else down.

ororopickpocket (changes requested):

PR #82, opened 2020-12-11
Fix tiny positions

jameshowlett977:


jameshowlett977:
Rollover questions:
LoanClosings
//TODO should we check total deposit here ?
//TODO loanLocal.borrower ?
//TODO should we skip this check if invoked from rollover ?
We need to double check these `todos`.
Changes should be done quickly if we need to do some.

contracts/modules/LoanClosings.sol
```diff
@@ -431,16 +441,43 @@ contract LoanClosings is LoanClosingsEvents, VaultController, InterestUser, Swap
...
-            loanParamsLocal.collateralToken,
-            loanLocal.principal,
-            loanLocal.collateral
-        );
-        require(
-            currentMargin > 3 ether, // ensure there's more than 3% margin remaining
-            "unhealthy position"
-        );
+        //close whole loan if tiny position will remain
+        //TODO should we check total deposit here ?
```
ororopickpocket, line 445: it's fine to take the principal / loan size everywhere

contracts/modules/LoanClosings.sol
```diff
@@ -466,6 +503,7 @@ contract LoanClosings is LoanClosingsEvents, VaultController, InterestUser, Swap
 
         Loan storage loanLocal = loans[loanId];
         LoanParams storage loanParamsLocal = loanParams[loanLocal.loanParamsId];
+        //TODO should we skip this check if invoked from rollover ?
```
ororopickpocket, line 506: it will probably cause the close to fail. did you test it?
the answer is: we should skip this when called from rollover.
david already extracted the check to the public depositWithCollateral function, I think. In that case, you would just need to merge his changes and it should work. However, his PR hasn't been reviewed, yet. So,  better wait a bit.

tjcloa (changes requested):
- verify all python tests are translated to hardhat then remove python tests
- resolve LoanTokenLogicStandard conflicts 
- check open comments if they are implemented and close them

contracts/modules/LoanClosingsBase.sol
```diff
@@ -356,32 +356,55 @@ contract LoanClosingsBase is LoanClosingsShared, LiquidationHelper {
...
 
 		if (rolloverReward != 0) {
-			// pay out reward to caller
-			loanLocal.collateral = loanLocal.collateral.sub(rolloverReward);
+			// if the reward > collateral:
+			if (rolloverReward > loanLocal.collateral) {
+				// 1. pay back the excess reward to the lender
+				// 2. pay the complete collateral to msg.sender
+				// 3. close the position & emit close event
+				uint256 excessRewards = rolloverReward.sub(loanLocal.collateral);
```
ororopickpocket, line 364: this does not make sense. if the rollover reqwrad is 100 and the collateral is 90, why would you just close 10? You need to close the complete thing.

contracts/modules/LoanClosingsBase.sol
```diff
@@ -356,32 +356,55 @@ contract LoanClosingsBase is LoanClosingsShared, LiquidationHelper {
...
+				_closeWithSwap(
+					loanLocal.id,
+					msg.sender,
+					excessRewards,
```
ororopickpocket, line 368: this should be the complete collateral

contracts/modules/LoanClosingsBase.sol
```diff
@@ -356,32 +356,55 @@ contract LoanClosingsBase is LoanClosingsShared, LiquidationHelper {
...
-				loanLocal.principal
-			);
-		} else {
-			(uint256 currentMargin, ) =
-				IPriceFeeds(priceFeeds).getCurrentMargin(
-					loanParamsLocal.loanToken,
-					loanParamsLocal.collateralToken,
-					loanLocal.principal,
-					loanLocal.collateral
+		if (loanLocal.collateral > 0) {
```
ororopickpocket, line 380: instead of checking the collateral here, you could just place the code below into the else block above - a sit is only to be executed in that context.

ororopickpocket (changes requested):

contracts/modules/LoanClosingsBase.sol
cwsnt: If we move to the else block above,
If let say the rolloverReward greater than 0, but smaller the collateral, the tiny position check won't be executed then?

contracts/modules/LoanClosingsBase.sol
```diff
@@ -361,11 +361,10 @@ contract LoanClosingsBase is LoanClosingsShared, LiquidationHelper {
 				// 1. pay back the excess reward to the lender
 				// 2. pay the complete collateral to msg.sender
 				// 3. close the position & emit close event
```
tjcloa, line 363: misleading comments not corresponding to the fixed logic

contracts/modules/LoanClosingsBase.sol
```diff
@@ -361,11 +361,10 @@ contract LoanClosingsBase is LoanClosingsShared, LiquidationHelper {
...
-				uint256 excessRewards = rolloverReward.sub(loanLocal.collateral);
 				_closeWithSwap(
 					loanLocal.id,
 					msg.sender,
-					excessRewards,
+					rolloverReward,
 					true, /// TODO::should we return it as the collateral or loan token?
```
tjcloa, line 368: is this TODO resolved?

contracts/modules/LoanClosingsBase.sol
```diff
@@ -361,11 +361,10 @@ contract LoanClosingsBase is LoanClosingsShared, LiquidationHelper {
 				// 1. pay back the excess reward to the lender
 				// 2. pay the complete collateral to msg.sender
 				// 3. close the position & emit close event
-				uint256 excessRewards = rolloverReward.sub(loanLocal.collateral);
 				_closeWithSwap(
 					loanLocal.id,
 					msg.sender,
-					excessRewards,
+					rolloverReward,
```
tjcloa, line 367: pls add test for this fix

contracts/modules/LoanClosingsBase.sol
cwsnt: Yes

contracts/modules/LoanClosingsBase.sol
cwsnt: This 3 comments is implemented in the `_closeWithSwap` function

ororopickpocket (changes requested):

contracts/modules/LoanClosingsRollover.sol
```diff
@@ -193,14 +193,14 @@ contract LoanClosingsRollover is LoanClosingsShared, LiquidationHelper {
...
+				// 1. pay back the remaining loan to the lender
+				// 2. pay the remaining collateral to msg.sender
 				// 3. close the position & emit close event
 				_closeWithSwap(
 					loanLocal.id,
 					msg.sender,
-					rolloverReward,
-					true,
+					loanLocal.collateral,
+					false,
```
eMarchenko, line 203: Changes in this contract look like fixes of a bug. Besides, testing should have detected it.

eMarchenko (commented):
looks good

PR #83, opened 2020-12-16
Change event

wukong-xxs:


contracts/modules/LoanOpenings.sol
```diff
@@ -355,7 +355,7 @@ contract LoanOpenings is LoanOpeningsEvents, VaultController, InterestUser, Swap
         );
 
         if (loanLocal.startTimestamp == block.timestamp) {
-            loanLocal.startRate = collateralToLoanRate;
+            loanLocal.startRate = isTorqueLoan ? collateralToLoanRate : sentValues[3];
```
ororopickpocket, line 358: there's one issue here. sentValues is not containing the collateralToLoanRate, but instead loanToCollateralRate.

so, we get different directions for different loans. that's not good. 

We need to decide for one direction. i suggest to convert sentValues accordingly, because we already have open loans on the contract and it would not be wise to change the direction now.

ororopickpocket (changes requested):

contracts/modules/LoanOpenings.sol
wukong-xxs: sentValues[3] is storing loanToCollateralSwapRate not loanToCollateralRate (they are different), so should we pass collateralToLoanSwapRate to loanLocal.startRate instead of  loanToCollateralSwapRate or collateralToLoanRate  when trading?

ororopickpocket: you need convert the sendValues[3] to the other direction. because then, we have consistent data on the contract.

ororopickpocket: it's the swap rate we need.

ororopickpocket (changes requested):

PR #87, opened 2020-12-18
Development vesting

jameshowlett977:


contracts/governance/Vesting/DevelopmentVesting.sol
```diff
@@ -0,0 +1,216 @@
...
+    function setTokenOwner(address _tokenOwner) public onlyOwner {
+        require(_tokenOwner != address(0), "token owner address invalid");
+        tokenOwner = _tokenOwner;
+    }
+
+    /**
+     * @notice send tokens to this contract, these tokens can be withdrawn any time by an owner
+     * @param _amount the amount of tokens to send
+     */
+    function sendTokens(uint _amount) public {
```
ororopickpocket, line 81: i find sendTokens a bit misleading. it sounds like we would send tokens from this smart contract to some address, although it's the other way round. Maybe rename to 'receiveTokens' or 'fetchTokens' or similar?

contracts/governance/Vesting/DevelopmentVesting.sol
```diff
@@ -0,0 +1,216 @@
...
+     */
+    function withdrawByGivenSchedules(address _receiver, uint _start, uint _number) public onlyOwners {
+        uint length = _start + _number;
+        if (length > schedules.length) {
+            length = schedules.length;
+        }
+        _withdrawBySchedules(_receiver, _start, length);
+    }
+
+    function _withdrawBySchedules(address _receiver, uint _start, uint _length) public {
```
ororopickpocket, line 173: why public?

contracts/governance/Vesting/DevelopmentVesting.sol
```diff
@@ -0,0 +1,216 @@
...
+        emit TokensWithdrawn(msg.sender, _receiver, _amount);
+    }
+
+    /**
+     * @notice change the vesting schedule
+     * @param _cliff the cliff in seconds
+     * @param _duration the total duration in seconds
+     * @param _frequency the withdrawal interval in seconds
+     */
+    function changeSchedule(uint _cliff, uint _duration, uint _frequency) public onlyOwner {
```
ororopickpocket, line 118: if i set schedule and cliff to 0, will everything unlock immediately?
(that'd be good to have, because in the future we might want to be able to change the vesting schedule only from the current time. we'd need to be able to unlock all tokens and send them to another contract, which could be done in a single governance proposal)

contracts/governance/Vesting/DevelopmentVesting.sol
```diff
@@ -0,0 +1,216 @@
...
+        require(success);
+
+        amount += _amount;
+
+        emit TokensSent(msg.sender, _amount);
+    }
+
+    //TODO onlyOwners or onlyOwner ?
+    /**
+     * @notice withdraws tokens
```
ororopickpocket, line 95: maybe add the information that only tokens which were not subject to vesting can be withdrawn by this function

jameshowlett977: done

ororopickpocket (approved):

contracts/governance/Vesting/DevelopmentVesting.sol
jameshowlett977: It was failed.
Changed, we can invoke `changeSchedule(0, 0, 0)` to unlock all tokens for all schedules. Test added.

contracts/governance/Vesting/DevelopmentVesting.sol
jameshowlett977: fixed

PR #90, opened 2020-12-25
Add SOV handler

wukong-xxs:


contracts/mixins/ProtocolTokenHandler.sol
```diff
@@ -0,0 +1,101 @@
...
+    mapping (uint256 => bool) usedNonces;
+    mapping (address => bool) verifiedSigners;
+
+    event AddSigner(address indexed singer);
+    event RemoveSigner(address indexed singer);
+    event SetRequiredCount(uint256 indexed requiredCount);
+    event Deposit(address indexed sender, uint256 indexed amount);
+    event Withdraw(address indexed caller, address indexed recipient, uint256 indexed amount);
+
+    function addSigner(address _signer) public onlyOwner {
```
ororopickpocket, line 24: good, but misses documentation

contracts/mixins/ProtocolTokenHandler.sol
```diff
@@ -0,0 +1,101 @@
...
+        require(_signer != address(0), "signer address should not be zero address");
+        require(!_signer.isContract(), "signer address should not be contract address");
+        require(!isSigner[_signer], "signer set");
+
+        isSigner[_signer] = true;
+
+        emit AddSigner(_signer);
+    }
+
+    function removeSigner(address _signer) public onlyOwner {
```
ororopickpocket, line 34: good, but misses documentation

contracts/mixins/ProtocolTokenHandler.sol
```diff
@@ -0,0 +1,101 @@
...
+        require(_signer != address(0), "signer address should not be zero address");
+        require(!_signer.isContract(), "signer address should not be contract address");
+        require(isSigner[_signer], "signer not set");
+
+        isSigner[_signer] = false;
+
+        emit RemoveSigner(_signer);
+    }
+
+    function setRequiredCount(uint256 requiredCount_) public onlyOwner {
```
ororopickpocket, line 44: same

contracts/mixins/ProtocolTokenHandler.sol
```diff
@@ -0,0 +1,101 @@
...
+        requiredCount = requiredCount_;
+
+        emit SetRequiredCount(requiredCount_);
+    }
+
+    /**
+     * @dev Before calling this function to deposit protocol tokens, 
+     * users need approve this contract to be able to spend or transfer their protocol tokens.    
+     */
+    function deposit(uint256 _amount) public {
```
ororopickpocket, line 56: @param missing in doc

contracts/mixins/ProtocolTokenHandler.sol
```diff
@@ -0,0 +1,101 @@
...
+
+        emit Deposit(msg.sender, _amount);
+    }
+
+    /**
+     * @param _data the message/data from users to be signed by authorized signers, should include in order: 
+     * address of recipient, addresss of destination contract (should be this contract address), amount of tokens to be withdrawed, nounce
+     * @param _signs array includes signs from authorized signers, amount/length should equal to requiredCount
+     */
+    function withdraw(bytes memory _data, bytes[] memory _signs) public {
```
ororopickpocket, line 70: would be better to pass ```address recipient, address contractAddress, uint256 amount``` instead of ```bytes data```. you can still hash it for the signature check

contracts/mixins/ProtocolTokenHandler.sol
```diff
@@ -0,0 +1,101 @@
...
+            require(isSigner[signer], "signer not authorized");
+            require(!verifiedSigners[signer], "signer verified");
+            verifiedSigners[signer] = true;
+        }
+
+        for (uint256 i = 0; i < requiredCount; i++) {
+            verifiedSigners[_hash.recover(_signs[i])] = false;
+        }
+
+        (address recipient, address contractAddress, uint256 amount, uint256 nonce) = abi.decode(_data, (address, address, uint256, uint256));
```
ororopickpocket, line 85: the nonce should not be passed on the function call.

contracts/mixins/ProtocolTokenHandler.sol
```diff
@@ -0,0 +1,101 @@
...
+
+        require(!usedNonces[nonce], "nonce used");
+        usedNonces[nonce] = true;
```
ororopickpocket, line 88: this allows the user to withdraw with an old nonce, if the signature hasn'T been used.

contracts/mixins/ProtocolTokenHandler.sol
```diff
@@ -0,0 +1,101 @@
...
+     * @param _data the message/data from users to be signed by authorized signers, should include in order: 
+     * address of recipient, addresss of destination contract (should be this contract address), amount of tokens to be withdrawed, nounce
+     * @param _signs array includes signs from authorized signers, amount/length should equal to requiredCount
+     */
+    function withdraw(bytes memory _data, bytes[] memory _signs) public {
+        _signs = new bytes[](requiredCount);
+        bytes32 _hash = keccak256(_data).toEthSignedMessageHash();
+
+        for (uint256 i = 0; i < requiredCount; i++) {
+            address signer = _hash.recover(_signs[i]);
```
ororopickpocket, line 75: i would have simply used ecrecover, because we do not rely on the uniqueness of a signature and we're adding a nonce. still, i appreciate your choice. it's very good practice

contracts/mixins/ProtocolTokenHandler.sol
```diff
@@ -0,0 +1,101 @@
...
+        emit Deposit(msg.sender, _amount);
+    }
+
+    /**
+     * @param _data the message/data from users to be signed by authorized signers, should include in order: 
+     * address of recipient, addresss of destination contract (should be this contract address), amount of tokens to be withdrawed, nounce
+     * @param _signs array includes signs from authorized signers, amount/length should equal to requiredCount
+     */
+    function withdraw(bytes memory _data, bytes[] memory _signs) public {
+        _signs = new bytes[](requiredCount);
```
ororopickpocket, line 71: aren't you here overriding the passed signatures?

contracts/mixins/ProtocolTokenHandler.sol
```diff
@@ -0,0 +1,101 @@
...
+        bytes32 _hash = keccak256(_data).toEthSignedMessageHash();
+
+        for (uint256 i = 0; i < requiredCount; i++) {
+            address signer = _hash.recover(_signs[i]);
+            require(isSigner[signer], "signer not authorized");
+            require(!verifiedSigners[signer], "signer verified");
+            verifiedSigners[signer] = true;
```
ororopickpocket, line 78: it's good you're checking for duplicate signatures. but you shouldn't use a storage variable for it. this is very gas expensive.

contracts/mixins/ProtocolTokenHandler.sol
ororopickpocket: the clean way is to have a nonce mapping per user address and to only process a nonce if it's higher than the last know nonce

contracts/mixins/ProtocolTokenHandler.sol
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        (address recipient, address contractAddress, uint256 amount, uint256 nonce) = abi.decode(_data, (address, address, uint256, uint256));
+
+        require(!usedNonces[nonce], "nonce used");
+        usedNonces[nonce] = true;
+
+        require(contractAddress == address(this), "contract address disagreement");
+
+        (address _protocolTokenAddress, bool success) = _withdrawProtocolToken(recipient, amount);
```
ororopickpocket, line 92: we'Re not withdrawing tokens from the protocol contract, but instead from this contract.

contracts/mixins/ProtocolTokenHandler.sol
```diff
@@ -0,0 +1,101 @@
+pragma solidity >=0.5.0 <0.6.0;
```
ororopickpocket, line 1: There are no units tests. We need unit tests for everything!

ororopickpocket (changes requested):

contracts/mixins/ProtocolTokenHandler.sol
wukong-xxs: Any better ideas?

contracts/mixins/ProtocolTokenHandler.sol
wukong-xxs: - what's the ProtocolTokenUser.sol for? 

- what's the difference and relationship between ProtocolTokenUser.sol and ProtocolTokenHandler.sol? 

- ProtocolTokenHandler.sol could inherit from State.sol (that means if it could use the variables of State like protocolTokenHeld, protocolTokenPaid, protocolTokenAddress)?

contracts/mixins/ProtocolTokenHandler.sol
wukong-xxs: here one issue, say, A wanna withdraw with nonce 1, B wanna withdraw with nonce 2, if B succeed to withdraw, that means A couldn't withdraw ever

contracts/mixins/ProtocolTokenHandler.sol
ororopickpocket: create a memory array and write the signer address to it. then check if a new signer is in the array.
it's more lines of code, but  more economic

contracts/mixins/ProtocolTokenHandler.sol
ororopickpocket: that's why i said 'per user address' :)

contracts/mixins/ProtocolTokenHandler.sol
ororopickpocket: ProtocolTokenUser is providing a method for withdrawing protocol tokens from the protocol itself.

This here (ProtocolTokenHandler) is not part of the prototocol. 
Instead, it's a isolated smart contract, which holds protocol tokens and lets users withdraw which have sufficient signatures. 
Therefore, it should not have any relationship with the ProtocolTokenUser.

Having ProtocolTokenHandler inherit from State.sol wouldn't have any benefit, because this contract is having its own storage. it's not part of the protocol.

contracts/mixins/ProtocolTokenHandler.sol
```diff
@@ -51,51 +65,75 @@ contract ProtocolTokenHandler is ProtocolTokenUser {
...
+        require(_amount > 0, "amount should larger than zero");
 
-        for (uint256 i = 0; i < requiredCount; i++) {
-            address signer = _hash.recover(_signs[i]);
-            require(isSigner[signer], "signer not authorized");
-            require(!verifiedSigners[signer], "signer verified");
-            verifiedSigners[signer] = true;
-        }
+        uint256 oldNonce = userNonce[msg.sender];
+        require(_nonce > oldNonce, "nonce smaller than last know nonce");
```
ororopickpocket, line 92: wukong, i'm sorry, because i requested this change, but i think your last solution was actually better fit for the purpose of this contract. could you rename nonce to taskID? and check that each taskID is only processed once per user? taskIDs can then be passed in any order. E.g. we can first process ID number 7, then ID number 3. but it has to be per user.

contracts/mixins/ProtocolTokenHandler.sol
```diff
@@ -51,51 +65,75 @@ contract ProtocolTokenHandler is ProtocolTokenUser {
...
-        }
+            sign = signs[i];
+            require(sign.length == 65, "length of sign is wrong");
 
-        (address recipient, address contractAddress, uint256 amount, uint256 nonce) = abi.decode(_data, (address, address, uint256, uint256));
+            assembly {
+                r := mload(add(sign, 0x20))
+                s := mload(add(sign, 0x40))
+                v := add(0x1b, byte(0, mload(add(sign, 0x60))))
+            }
```
ororopickpocket, line 115: that's a way to do it and i guess figuring that out was not easy. i'm glad you learned how to access calldata using assembly, but you could have made your life easier and just pass  uint8[] v, bytes32[] r and bytes32[] s.

This is just a comment to let you know this would have been an option. you can leave it as it is ;)

contracts/mixins/ProtocolTokenHandler.sol
```diff
@@ -51,51 +65,75 @@ contract ProtocolTokenHandler is ProtocolTokenUser {
...
+            verifiedSigners[i] = signer;
+        }
 
-        (address _protocolTokenAddress, bool success) = _withdrawProtocolToken(recipient, amount);
-        require(_protocolTokenAddress == protocolTokenAddress, "protocol token address disagreement");
-        if (success) {
-            protocolTokenPaid = protocolTokenPaid.add(amount);
+        uint256 balance = IERC20(protocolTokenAddress).balanceOf(address(this));
+        if (_amount > balance) {
+            _amount = balance;
```
ororopickpocket, line 129: good thinking, but i think, it would actually be better to revert if the amount is bigger than the balance. the user will expect to get the full reward and not just part of it. it's better to wait for a contract refill instead of making partial payments.

ororopickpocket (changes requested):

ororopickpocket:
note: this contract is to be used for paying out SOV rewards to users. Users whould be able to claim their rewards themselves if they can prove that it's approved by sufficient signers.

ororopickpocket (approved):

contracts/mixins/ProtocolTokenHandler.sol
```diff
@@ -0,0 +1,138 @@
...
+        isSigner[_signer] = false;
+
+        emit RemoveSigner(_signer);
+    }
+
+    /**
+     * @param requiredCount_ required amount of signs provided by users
+     */
+    function setRequiredCount(uint256 requiredCount_) public onlyOwner {
+        require(requiredCount > 0, "required amount of signs should larger than zero");
```
eMarchenko, line 59: shouldn't it be `requiredCount_`?

contracts/mixins/ProtocolTokenHandler.sol
```diff
@@ -0,0 +1,138 @@
...
+            sign = signs[i];
+            require(sign.length == 65, "length of sign is wrong");
+
+            assembly {
+                r := mload(add(sign, 0x20))
+                s := mload(add(sign, 0x40))
+                v := add(0x1b, byte(0, mload(add(sign, 0x60))))
+            }
+
+            signer = ecrecover(_hash, v, r, s);
```
eMarchenko, line 120: just use `ECDSA.recover`. it has some extra checks and is bug-free

contracts/mixins/ProtocolTokenHandler.sol
```diff
@@ -0,0 +1,138 @@
...
+        
+        uint8 v;
+        bytes32 r;
+        bytes32 s;
+        address signer;
+        bytes memory sign;
+        bytes[] memory signs = new bytes[](requiredCount);
+        address[] memory verifiedSigners = new address[](requiredCount);
+
+        signs = _signs;
```
eMarchenko, line 108: I guess it was used to convert `calldata` to `memory`. but you have `public` function with `memory` parameter, so this extra conversion only burns gas: `calldata` argument is already copies to `memory` internally.

contracts/mixins/ProtocolTokenHandler.sol
```diff
@@ -0,0 +1,138 @@
...
+
+    constructor (address protocolTokenAddress_) public {
+        require(protocolTokenAddress_.isContract(), "protocol token address should be contract address");
+        protocolTokenAddress = protocolTokenAddress_;
+    }
+
+    /** 
+     * @param _signer address of signer added
+     */
+    function addSigner(address _signer) public onlyOwner {
```
eMarchenko, line 32: consider marking functions `external` if possible (multiple instances in this PR). it saves a bit of gas and improves readability.

contracts/mixins/ProtocolTokenHandler.sol
```diff
@@ -0,0 +1,138 @@
...
+                s := mload(add(sign, 0x40))
+                v := add(0x1b, byte(0, mload(add(sign, 0x60))))
+            }
+
+            signer = ecrecover(_hash, v, r, s);
+            require(isSigner[signer], "signer not authorized");
+
+            for (uint256 q = 0; q < i; q++) {
+                require(signer != verifiedSigners[q], "signer verified");
+            }
```
eMarchenko, line 125: you could sort signatures off-chain so that `verifiedSigners` will be a sorted array. there is little difference if the array is short, of course.

contracts/mixins/ProtocolTokenHandler.sol
```diff
@@ -0,0 +1,138 @@
...
+    address public protocolTokenAddress;
+    uint256 public requiredCount = 2;
+    mapping (address => bool) public isSigner;
+    mapping (address => uint256) public userNonce;
+
+    event AddSigner(address indexed signer);
+    event RemoveSigner(address indexed signer);
+    event SetRequiredCount(uint256 indexed requiredCount);
+    event Deposit(address indexed sender, uint256 indexed amount);
+    event Withdraw(address indexed caller, address indexed recipient, uint256 indexed amount);
```
eMarchenko, line 22: `uint` parameters of events should rarely be `indexed`. `address` parameters are mostly indexed.
think of it as if "do you want to find all events with this particular argument value?".
every indexed parameter costs you a bit of extra gas

eMarchenko (changes requested):
The code is ok except for several low-severity issues:
- a bug with minimal impact
- you should use ECDSA library when possible
- some minor gas improvements

contracts/mixins/ProtocolTokenHandler.sol
wukong-xxs: https://github.com/DistributedCollective/Sovryn-smart-contracts/pull/90#discussion_r549722615

eMarchenko: oh.

contracts/openzeppelin/ECDSA.sol
```diff
@@ -1,4 +1,4 @@
-pragma solidity ^0.5.0;
+pragma solidity >=0.5.0 <0.6.0;
```
eMarchenko, line 1: these two variants are equal, and `^` notation is just more popular.

eMarchenko (approved):

PR #91, opened 2020-12-29
governance: fixes and test for cancelation

jameshowlett977:


jameshowlett977:
We need to answer on question: how to change checking?
I see at least 3 variants:
- nobody has voted for the proposal
- voting hasn't been started yet
- the proposal hasn't been queued yet

ororopickpocket (approved):

PR #92, opened 2020-12-30
Governance multisig keyholders

jameshowlett977:


ororopickpocket (changes requested):
Quality work like expected :)
There's just one thing missing (which i think i didn't specify), but i think the governance should also be able to change the required number of signatures for the ETH and BTC multisig.

jameshowlett977:
Thanks :-)
Done

ororopickpocket (approved):

PR #94, opened 2020-12-31
Governance test

powerhousefrank:
Included Tests:

- User Type: Proposers

    * Should not create a new proposal if not enough staked
    
- User Type: Voter

    * Should not be allowed to vote on a proposal with any other state than active
    * Voting should emit the VoteCast Event

- User Type: Guardian

    * Remove a successful proposal which was queued even if successful
    * Cannot remove a proposal which is already executed
    * Make a new Governor the admin of current Timelock with being pendingAdmin
    * Should not be able to make a new Governor the admin of current Timelock without being pendingAdmin first    
    * Should be possible to abdicate being a guardian
    * Should not be able to perform any task related to guardian after abdication

- User Type: Anyone
    * Can queue a successful proposal
    * Cannot queue a defeated proposal
    * Adding a proposal to queue should emit ProposalQueued event
    * Can execute a queued proposal
    * Cannot execute a proposal which is not queued
    * Cannot execute a proposal which is already executed
    * All actions mentioned in the queue of a proposal should be executed correctly
    * Executing a proposal should emit the ProposalExecuted Event
    * Cannot remove a proposal which is already executed

### DIFF
--- a/contracts/mockup/setGet.sol
+++ b/contracts/mockup/setGet.sol
@@ -1,0 +1,28 @@
+pragma solidity 0.5.17;
+
+/**
+ * @title A simple smart contract for setting and getting values.
+ * @author Franklin Richards
+ * @dev This is going to be used for testing purposes.
+ */
+contract setGet {
+    uint256 public value;
+
+    event valueSet(uint256 indexed _value);
+
+    /**
+     * @notice To get the `value`.
+     * @return _value The value.
+     */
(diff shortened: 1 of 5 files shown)


ororopickpocket (commented):
very good work. just one request: please, remove the empty test cases which are present in other files. you can have a comment explaining which test cases would be there if they were not yet written in other files, but i would prefer them to not show up when calling 'truffle test'.

PR #100, opened 2021-01-07
Adding CI to the development branch

powerhousefrank:
Things included in the PR:
- prettier (For beautification of the Solidity and JS code)
- solhint or lint-contracts (For linting)
- coveralls (For code coverage)

The above three would be running in Travis, each time you make a push to a branch or create a new PR.

- slither (For static analysis)
- brownie (For python test)

The above two would be running in your local machine if you have installed all the dependencies as mentioned in the updated Readme in this commit.

Each time you commit, it will run contract static analysis, prettier and linting.

Each time you create a push, it will run all the test cases (both python and JS, using brownie for python tests and truffle for JS tests)

If for some reason, you want to skip these checks (NOT RECOMMENDED), then you could add `--no-verify` at the end of the git command to skip these checks.

There are many linting and prettier changes to the files, apart from LoanTokenStandardLogic.sol (contracts/connectors/loantoken), there should not be any logical changes. (Please verify the LoanTokenStandardLogic.sol change as well. Ternary Operation has been flattened.)


.prettierrc
```diff
@@ -0,0 +1,15 @@
+{
+  "overrides": [
+    {
+      "files": "*.sol *.js",
+      "options": {
+        "printWidth": 120,
```
jameshowlett977, line 6: I think 140 is better.

.solhint.json
```diff
@@ -2,6 +2,7 @@
   "extends": "solhint:default",
   "plugins": ["prettier"],
   "rules": {
-    "prettier/prettier": "error"
+    "prettier/prettier": "error",
+    "max-line-length": ["warn",120]
```
jameshowlett977, line 6: I think 140 is better.

jameshowlett977 (approved):

.solhint.json
powerhousefrank: Updated to 140 in the latest commit aa878fd

.prettierrc
powerhousefrank: Updated to 140 in the latest commit aa878fd

ororopickpocket (approved):
looks good. please, resolve the merge conflicts and merge into development

ororopickpocket:
oh, one note: the linter is still expecting a line width of 120 instead of 140

powerhousefrank:
> oh, one note: the linter is still expecting a line width of 120 instead of 140

If it is from the Travis, I have cleared the cache, now it's checking for 140 only.

PR #101, opened 2021-01-08
Add current price to the DepositCollateral event

wukong-xxs:


contracts/modules/LoanMaintenance.sol
```diff
@@ -90,8 +90,15 @@ contract LoanMaintenance is LoanOpeningsEvents, LoanMaintenanceEvents, VaultCont
                 msg.value
             );
         }
+
+        (, uint256 collateralToLoanRate) = IPriceFeeds(priceFeeds).getCurrentMargin(
+            loanParamsLocal.loanToken,
+            loanParamsLocal.collateralToken,
+            loanLocal.principal,
+            loanLocal.collateral
+        );
```
ororopickpocket, line 100: we don't need to compute the margin here. why don't you just call ```queryRate```? Would be cheaper

ororopickpocket (changes requested):

ororopickpocket (approved):

PR #103, opened 2021-01-11
Temporary governance

jameshowlett977:
Draft PR to review changes


contracts/governance/comp/GovernorVault.sol
```diff
@@ -0,0 +1,52 @@
...
+     * @notice transfers tokens
+     * @param _receiver the receiver of tokens
+     * @param _amount the to be transferred
+     */
+    function transferTokens(address _receiver, uint _amount) public onlyOwner {
+        require(_receiver != address(0), ERROR_INVALID_ADDRESS);
+
+        require(token.transfer(_receiver, _amount), "Transfer failed");
+        emit TokensTransferred(_receiver, _amount);
+    }
```
ororopickpocket, line 32: this will work for this purpose. for the actual governance i would prefer to not be limited to a single token contract, but instead just pass the token address to the transferToken function

jameshowlett977: fixed

PR #105, opened 2021-01-15
Add divergence check and max slippage

wukong-xxs:


ororopickpocket (changes requested):
It looks good, Wukong, but we'll need some changes, nonetheless. 

The frontend is preparing a confirmation window where the user can select the maximum slippage he's allowing from the expected price he's being shown. However, this will not be the oracle price (like it is the case now on the trading page), but the expected price from the AMM. This means, a divergence check between AMM and oracle price is not what we need.

What we need instead is to allow the user to pass the minReturn value, similar to how it is on the oracle based amm. In case of the swap, you can simply pass on the value to the AMM contracts. they will handle it automatically. in case of the margin trade it's a bit more complicated. 

one option is to let the user pass the minimum position size in the collateral currency. 

there are certainly other options as well. feel free to propose one if you find another one better suited.

tests/test_SwapsExternal.py
```diff
@@ -0,0 +1,54 @@
...
+'''
+def test_swapExternal_with_larger_minReturn(accounts, sovryn, RBTC, SUSD, priceFeeds):
+    SUSD.approve(sovryn.address, 1e18)
+    with reverts("destTokenAmountReceived too low"):
+        sovryn.swapExternal(SUSD.address, RBTC.address, accounts[0], accounts[0], 1e18, 0, 1e18, b'', priceFeeds.address)
+
+
+def test_getSwapExpectedReturn(accounts, sovryn, RBTC, SUSD):
+    result = sovryn.getSwapExpectedReturn(SUSD.address, RBTC.address, 1e18)
+    assert(result > 0)
```
ororopickpocket, line 49: better would be to check the actual number which is expected

tests/test_SwapsExternal.py
```diff
@@ -0,0 +1,54 @@
...
+
+    tx = sovryn.swapExternal(SUSD.address, RBTC.address, accounts[0], accounts[0], 1e18, 0, 0, b'', priceFeeds.address)
+    
+    external_swap = tx.events['ExternalSwap']
+    assert(external_swap['user'] == accounts[0])
+    assert(external_swap['sourceToken'] == SUSD.address)
+    assert(external_swap['destToken'] == RBTC.address)
+    assert(external_swap['sourceAmount'] == 1e18)
+    assert(external_swap['sourceAmount'] == tx.return_value[1])
+    assert(external_swap['destAmount'] == tx.return_value[0])
```
ororopickpocket, line 17: better to also check the actual value. it should be the same as the expected one.

contracts/connectors/loantoken/LoanTokenLogicStandard.sol
```diff
@@ -215,74 +215,70 @@ contract LoanTokenLogicStandard is LoanTokenSettingsLowerAdmin {
...
+				loanId,
+				0, // withdrawAmount
+				leverageAmount,
+				collateralTokenAddress,
+				sentAddresses,
+				sentAmounts,
+				loanDataBytes
+			);
+
+		require(newCollateral >= minReturn, "new collateral too low");
```
ororopickpocket, line 281: if you're having the check here, the user has to pay the full amount if gas even if his transaction fails. it's better to have it at the beginning and use the same function as the SwapsExternal (see comments there)

contracts/modules/SwapsExternal.sol
```diff
@@ -75,6 +80,8 @@ contract SwapsExternal is VaultController, SwapsUser {
             swapData
         );
 
+        require(destTokenAmountReceived >= minReturn, "destTokenAmountReceived too low");
```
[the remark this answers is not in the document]
ororopickpocket: you're missing unit tests for this. make a realistic test case, please.

rate on the contract is 1 BTC = 10000 USD
let's assume you do not want to lose more than 2%. Calculate the expected collateral size, pass it to the function and change the rate on the oracle. See if it's behaving as expected.

ororopickpocket (changes requested):

contracts/connectors/loantoken/LoanTokenLogicDai.sol
```diff
@@ -185,6 +185,7 @@ contract LoanTokenLogicDai is LoanTokenLogicStandard {
         uint256 collateralTokenSent,
         address collateralToken,
         address trader,
+		uint256 minReturn,
```
eMarchenko, line 188: some `tab` issues

contracts/connectors/loantoken/LoanTokenLogicStandard.sol
```diff
@@ -215,74 +215,70 @@ contract LoanTokenLogicStandard is LoanTokenSettingsLowerAdmin {
...
-            _borrowOrTrade(
-                loanId,
-                0, // withdrawAmount
-                leverageAmount,
-                collateralTokenAddress,
-                sentAddresses,
-                sentAmounts,
-                loanDataBytes
-            );
+		require(minReturn >= 0, "minReturn must larger than zero");
```
eMarchenko, line 231: it is `uint`, it is always greater or equal zero.  most probably, should be `require(minReturn > 0, ...)`

contracts/connectors/loantoken/LoanTokenLogicStandard.sol
```diff
@@ -215,74 +215,70 @@ contract LoanTokenLogicStandard is LoanTokenSettingsLowerAdmin {
...
+
+		if (collateralTokenAddress == address(0)) {
+			collateralTokenAddress = wrbtcTokenAddress;
+		}
+
+		require(collateralTokenAddress != loanTokenAddress, "11");
+
+		//temporary: limit transaction size
+		if (transactionLimit[collateralTokenAddress] > 0) require(collateralTokenSent <= transactionLimit[collateralTokenAddress]);
```
eMarchenko, line 240: `require(transactionLimit[collateralTokenAddress] == 0 || collateralTokenSent <= transactionLimit[collateralTokenAddress]);` is arguably more readable, especially
```ts 
uint limit = transactionLimit[collateralTokenAddress];
require(limit == 0 || collateralTokenSent <= limit);
```

contracts/modules/SwapsExternal.sol
```diff
@@ -30,15 +31,19 @@ contract SwapsExternal is VaultController, SwapsUser {
...
+        address priceFeeds_
     )
         external
         payable
         nonReentrant
         returns (uint256 destTokenAmountReceived, uint256 sourceTokenAmountUsed)
     {
         require(sourceTokenAmount != 0, "sourceTokenAmount == 0");
-
+        require(minReturn >= 0, "minReturn must larger than zero");
```
eMarchenko, line 44: `minReturn > 0`

contracts/modules/SwapsExternal.sol
```diff
@@ -91,4 +98,18 @@ contract SwapsExternal is VaultController, SwapsUser {
...
+        address destToken,
+        uint256 sourceTokenAmount,
+        address priceFeeds_) 
+        public 
+        returns(bool result) 
+    {
+        (uint256 rate, uint256 precision) = IPriceFeeds(priceFeeds_).queryRate(sourceToken, destToken);
+        uint256 destTokenAmount = _swapsExpectedReturn(sourceToken, destToken, sourceTokenAmount);
+        uint256 rateFromSwap = destTokenAmount.mul(precision).div(sourceTokenAmount);
+        result = rate == rateFromSwap ? true : false;
```
eMarchenko, line 113: just `result = rate == rateFromSwap;`
using `true` and `false` in comparisons and other logical expressions is generally discouraged

contracts/modules/SwapsExternal.sol
```diff
@@ -30,15 +31,19 @@ contract SwapsExternal is VaultController, SwapsUser {
         address returnToSender,
         uint256 sourceTokenAmount,
         uint256 requiredDestTokenAmount,
-        bytes calldata swapData
+        uint256 minReturn,
+        bytes calldata swapData,
+        address priceFeeds_
```
eMarchenko, line 36: why the caller controls `priceFeeds_`? it looks dangerous. Feels like oracle manipulation for free.

contracts/modules/SwapsExternal.sol
```diff
@@ -91,4 +98,18 @@ contract SwapsExternal is VaultController, SwapsUser {
     ) external view returns (uint256) {
         return _swapsExpectedReturn(sourceToken, destToken, sourceTokenAmount);
     }
+
+    function checkPriceDivergence(        
+        address sourceToken,
+        address destToken,
+        uint256 sourceTokenAmount,
+        address priceFeeds_)
```
eMarchenko, line 106: `priceFeeds_` controlled by the caller

eMarchenko (changes requested):
`checkPriceDivergence` and `priceFeeds_` issues probably deserve additional discussion

interfaces/ISovryn.sol
```diff
@@ -0,0 +1,357 @@
...
+import "../contracts/core/State.sol";
+import "../contracts/events/ProtocolSettingsEvents.sol";
+import "../contracts/events/LoanSettingsEvents.sol";
+import "../contracts/events/LoanOpeningsEvents.sol";
+import "../contracts/events/LoanMaintenanceEvents.sol";
+import "../contracts/events/LoanClosingsEvents.sol";
+import "../contracts/events/FeesEvents.sol";
+import "../contracts/events/SwapsEvents.sol";
+
+contract ISovryn is
```
jameshowlett977, line 18: Why do we need second ISovryn interface?

jameshowlett977 (commented):
Please add tests for the following functions:
checkPriceDivergence
marginTrade (minReturn/minPoisitonSize > 0)
swapExternal

Passing minReturn/minPoisitonSize = 0 to existing tests is not enough.

korepkorep (commented):

PR #129, opened 2021-01-28
Affiliates

tjcloa:
Affiliates done. 100% coverage, events covered as well. 
Also fixed js tests issues due to migration to hardhat and set correct objects where LoanClosings was used: now it is split into LoanClosingsBase and LoanClosingsWith, both behind the proxy, but needs to be applied to other branches that had been checked out before.


tjcloa:
moved over to https://github.com/DistributedCollective/Sovryn-smart-contracts/tree/affiliates_v2

PR #130, opened 2021-02-01
reintroduce whitelisting

ororopickpocket:


contracts/connectors/loantoken/LoanTokenSettingsLowerAdmin.sol
```diff
@@ -22,8 +22,20 @@ contract LoanTokenSettingsLowerAdmin is AdvancedToken {
 	// ------------- END MUST BE THE SAME AS IN LoanToken CONTRACT -------------------
 
 	//Add new variables here on the bottom
+	address public earlyAccessToken;
```
ororopickpocket, line 25: note: a version with early access token was already deployed in the past, before the pauser was added. no version with pauser has been deployed, yet. hence, the order.

tjcloa (approved):

PR #139, opened 2021-02-10
Fix position size calculation

wukong-xxs:


ororopickpocket:
wukong, it seems you started from an incorrect branch

ororopickpocket:
you've got an unclean commit history, which franklin cleaned days ago

PR #140, opened 2021-02-10
Vesting optimization

jameshowlett977:
- added VestingLogic, will be deployed once
- split Proxy on Proxy and UpgradableProxy
- Vesting, TeamVesting are simple proxies without upgradability (Proxy is a base contract)
- Vesting implements function governanceWithdrawTokens with simple revert to disable this feature


ororopickpocket:
got it. you're overwriting the governanceWithdrawTokens

jameshowlett977:
> got it. you're overwriting the governanceWithdrawTokens

Yes,
The difference of these 2 vesting contracts is only one function `governanceWithdrawTokens`.
Vesting shouldn't support it
TeamVesting should support it
I've disabled this feature for Vesting:
`function governanceWithdrawTokens(address receiver) public {
		revert("operation not supported");
	}`
Had to change inherinace order, otherwise it will be disabled for TeamVesting as well.

ororopickpocket:
One request and one remark.

I think, we should add a function to change the vesting logic address on the vesting factory, should we ever decide to deploy new vesting contracts with an updated logic.

I think, we should set the proxy owner to 0 for all vesting contracts which are not under control of the team multisig, because neither the team nor the governance should be able to adapt investor vesting contracts. Neither do we want to allow the investor changing the logic himself.

jameshowlett977:
> I think, we should add a function to change the vesting logic address on the vesting factory, should we ever decide to deploy new vesting contracts with an updated logic.

Added setVestingFactory to VestingRegistry.
We will be able to redeploy VestingFactory and VestingLoigc,
functions of VestingFactory can be invoked only by VestingRegistry.

> I think, we should set the proxy owner to 0 for all vesting contracts which are not under control of the team multisig, because neither the team nor the governance should be able to adapt investor vesting contracts. Neither do we want to allow the investor changing the logic himself.

Vesting and TeamVesting contract doesn't have function setImplementation.
I've added `_setProxyOwner(address(this));` to mark explicidly than noone will be able to invoke it even if contracts inherited from UpgradableProxy accidentally.

contracts/proxy/UpgradableProxy.sol
```diff
@@ -0,0 +1,24 @@
+pragma solidity ^0.5.17;
```
eMarchenko, line 1: Both `Proxy.sol` and `UpgradableProxy.sol` are slightly different from their [OZ counterparts](https://github.com/OpenZeppelin/openzeppelin-contracts/tree/master/contracts/proxy). 
I'd recommend using their proxy unless there is a strong argument against it (like your contracts are already deployed and are incompatible on storage layout level)

contracts/governance/Vesting/TeamVesting.sol
```diff
@@ -1,29 +1,51 @@
 pragma solidity ^0.5.17;
 pragma experimental ABIEncoderV2;
 
-import "./Vesting.sol";
+import "../../openzeppelin/Ownable.sol";
```
eMarchenko, line 4: I'd recommend to manage OZ code as a dependency (via NPM). 
btw, the project doesn't include `package-lock.json`...
these notes are arbitrary placed in this PR (could be in any other)

contracts/governance/Vesting/TeamVesting.sol
```diff
@@ -1,29 +1,51 @@
...
+	) public {
+		require(_SOV != address(0), "SOV address invalid");
+		require(_stakingAddress != address(0), "staking address invalid");
+		require(_tokenOwner != address(0), "token owner address invalid");
+		require(_duration >= _cliff, "duration must be bigger than or equal to the cliff");
+		require(_feeSharingProxy != address(0), "feeSharingProxy address invalid");
 
-		_withdrawTokens(receiver, true);
+		_setImplementation(_logic);
+		_setProxyOwner(address(this));
```
eMarchenko, line 42: It's a nice touch, though it mostly burns gas (proxy Owner doesn't have any functions anyway)

contracts/governance/Vesting/VestingStorage.sol
```diff
@@ -0,0 +1,27 @@
+pragma solidity ^0.5.17;
+
+import "../../openzeppelin/Ownable.sol";
+import "../../interfaces/IERC20.sol";
+import "../Staking/Staking.sol";
+import "../IFeeSharingProxy.sol";
+
+contract VestingStorage is Ownable {
```
eMarchenko, line 8: do you keep this `Ownable` for preserve the storage layout?
you might want to add a comment about it, otherwise it's counterintuitive

eMarchenko (approved):
the code is good
I've added some general comments that mostly shouldn't be addressed in this PR

contracts/governance/Vesting/TeamVesting.sol
jameshowlett977: agree, added it in a rush :-)
removed

contracts/governance/Vesting/VestingStorage.sol
jameshowlett977: yes, comment added

contracts/governance/Vesting/TeamVesting.sol
jameshowlett977: I've added OZ npm package initially,
but we have contracts in `openzeppelin` folder historically and
working together with these contracts and contracts from npm package is also not so good.
So we agreed to use contracts from `openzeppelin` folder.

contracts/proxy/UpgradableProxy.sol
jameshowlett977: I would agree with you if not tomorrow's deployment.

eMarchenko (approved):

ororopickpocket (approved):

powerhousefrank (approved):

PR #142, opened 2021-02-10
Add internalExpectedReturn to fix the position size calculation

wukong-xxs:


ororopickpocket (changes requested):
looks good, but please extend the tests.
the margin trade test where you added the print statements should verify the numbers of the trade event
i think for the estimatedMarginDetails there's also a separate test. if yes, it probably needs adaption. If not, please create one

ororopickpocket (approved):

Issue #138, opened 2021-02-10
Development Fund Deployment Enhancement


powerhousefrank:
Currently to deploy the development fund, it would require atleast 2 days to pass the release schedule through governance.

Solution:
- To use an init() which could be only called once after contract deployment which will be used to fund tokens.

powerhousefrank:
Solved in #143

PR #143, opened 2021-02-10
Development Fund Contract & Test Updated

powerhousefrank:
This solves the issue #138

### DIFF
--- a/contracts/governance/Vesting/DevelopmentFund.sol
+++ b/contracts/governance/Vesting/DevelopmentFund.sol
@@ -16,6 +16,10 @@ contract DevelopmentFund {
 	/// @notice The SOV token contract.
 	IERC20 public SOV;
 
+	/// @notice The current contract status.
+	enum Status {Deployed, Active, Expired}
+	Status public status;
+
 	/// @notice The owner of the locked tokens (usually Governance).
 	address public lockedTokenOwner;
 	/// @notice The owner of the unlocked tokens (usually MultiSig).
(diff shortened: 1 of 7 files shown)


contracts/governance/Vesting/DevelopmentFund.sol
```diff
@@ -121,14 +143,47 @@ contract DevelopmentFund {
...
+		/// Checking if the schedule duration and token allocation length matches.
+		require(_releaseDuration.length == _releaseTokenAmount.length, "Release Schedule does not match.");
+
+		/// Finally we update the token release schedule.
+		releaseDuration = _releaseDuration;
+		releaseTokenAmount = _releaseTokenAmount;
+	}
+
+	/**
+	 * @notice This function is called once after deployment for token transfer based on schedule.
```
eMarchenko, line 161: this solution works, though it's not so elegant.
Alternatively you could precompute contract's address before deployment, transfer tokens to this address while it has no  bytecode, and verify inside constructor that the contract has the proper SOV balance. This approach saves some Solidity code, but puts extra stress on your deployment procedure.

eMarchenko (approved):
The update is well implemented according to the chosen design.

contracts/governance/Vesting/DevelopmentFund.sol
powerhousefrank: I don't think I want to stress @jameshowlett977 more than he is already. I think we can take this route.

jameshowlett977 (approved):

PR #145, opened 2021-02-15
Governance Voting contract for different options. reads the voting power from staking contract.

JohnAllerdyce:
The voting contract for different options. reads the voting power from staking contract.

- proposal isn't executable
- dummy/fake data can be used for targets, signatures, calldatas
- we don't need to change anything related to functions cancel, queue, execute
- Timelock remains the same
- dynamic number of options during proposal creation
- the users just choose between options defined

- we need a list of options for the proposal, can be dynamic array of string (bytes32 can be considered, but could be not enough) instead of bool 
struct Receipt {
        bool hasVoted;
        uint8 option; (can be uint16, but I think 256 options is enough)
        uint96 votes;
    }

'struct Proposal {
        uint256 id;
        uint32 startBlock;
        uint32 endBlock;
        //uint96 forVotes;
        //uint96 againstVotes;
                mapping (uint8 => uint96) votes;'


JohnAllerdyce:
converting to draft

JohnAllerdyce:
converting to draft

PR #146, opened 2021-02-19
Origin investors claim vested sov

tjcloa:
Note: 
1. the contract needs to be funded with SOV tokens right after deployment, either transferred SOV and set allowance
2. vestingRegistry.sol addAdmin(the contract address) should be called to allow creating vesting contracts


contracts/governance/Vesting/OriginInvestorsClaim.sol
```diff
@@ -80,46 +68,83 @@ contract OriginInvestorsClaim is Ownable {
 	}
 
 	//in case we have unclaimed tokens
-	function ownerTtransferBalance(address toAddress) public onlyOwner {
-		SOVToken.transfer(toAddress, SOVToken.balanceOf(address(this)));
+	function authorizedBalanceWithdraw(address toAddress) public onlyAuthorized {
+		require(
+			SOVToken.transfer(toAddress, SOVToken.balanceOf(address(this))),
```
eMarchenko, line 73: should check return value (erc20 standard `MUST` requirement)

contracts/governance/Vesting/OriginInvestorsClaim.sol
```diff
@@ -80,46 +68,83 @@ contract OriginInvestorsClaim is Ownable {
...
-	function createVesting() internal notInVestingList {
+	function createVesting() internal {
 		uint256 cliff = vestingTerm.sub(now);
 		uint256 duration = cliff;
 		uint256 amount = investorsAmountsList[msg.sender];
 		address vestingContractAddress;
 
-		vestingList[msg.sender] = true;
+		vestingContractAddress = vestingRegistry.getVesting(msg.sender);
+		require(vestingContractAddress == address(0), "OriginInvestorsClaim::withdraw: investor has an active vesting contract");
```
eMarchenko, line 126: are you sure that an investor cannot have another vesting already? from "some other source"

eMarchenko (commented):
consider checking return values of ERC20 calls all over the code

contracts/governance/Vesting/OriginInvestorsClaim.sol
ororopickpocket: excellent point. we need to consider that a lot of origin investors already invested in genesis. so, that will likely be the case for some. these people still need to be able to claim their tokens. therefore, they should be able to pass an address which should be the token owner.

ororopickpocket: this also means that we need to communicate to our users that they HAVE to pass another address if they invested in genesis. victor needs to know that the frontend should check if the user is cSOV holder.

contracts/governance/Vesting/OriginInvestorsClaim.sol
```diff
@@ -0,0 +1,150 @@
...
+		uint256[] memory claimAmounts,
+		address allowanceFrom
+	) public onlyAuthorized {
+		uint256 totalAmount;
+		require(
+			investors.length == claimAmounts.length,
+			"OriginInvestorsClaim::setInvestorsAmountsList: investors.length != claimAmounts.length"
+		);
+		require(!investorsListInitialized, "OriginInvestorsClaim::setInvestorsAmountsList: the investors list has already been set");
+		investorsListInitialized = true;
```
ororopickpocket, line 95: this means we'll be able to call the function only once. We've got over 3000 genesis investors. I doubt we can set all of them in a single transaction considering the RSK gas limit is only 6.8M.

ororopickpocket (changes requested):
A few general comments apart from the change request:

* All new feature branches should branch from development, not from other, unmerged branches
* All function should be documented

contracts/governance/Vesting/OriginInvestorsClaim.sol
ororopickpocket: I looked up the numbers. we've got 3284 investors.

ororopickpocket: how many can we process in a single tx?

contracts/governance/Vesting/OriginInvestorsClaim.sol
eMarchenko: I haven't dived deep into the vesting, however I have a feeling that it is possible to retrieve an existing vesting contract and add more funds there (i.e. reuse existing user's vesting contract for origin vesting). Initial code's logic was close to that behavior.

contracts/governance/Vesting/OriginInvestorsClaim.sol
eMarchenko: it takes one `sstore` per investor (20k gas), which translates to about 300 investors per transaction

ororopickpocket:
code was merged with another PR

PR #152, opened 2021-02-25
Sip 006a and 008 deployment script

powerhousefrank:
SIPs included:

- [x] https://github.com/DistributedCollective/SIPS/blob/9203633/SIP-0006(A1).md
- [x] https://github.com/DistributedCollective/SIPS/blob/a8cf098/SIP-0008.md

### DIFF
--- a/scripts/contractInteraction/contract_interaction.py
+++ b/scripts/contractInteraction/contract_interaction.py
@@ -113,20 +113,18 @@ def main():
     # setEarlyAccessToken(contracts['iRBTC'], contracts['og'])
     # setEarlyAccessToken(contracts['iBPro'], contracts['og'])
 
-    createProposalSIP005()
+    # createProposalSIP005()
+    checkVotingPower()
 
 def loadConfig():
     global contracts, acct
     this_network = network.show_active()
     if this_network == "rsk-mainnet":
         configFile =  open('./scripts/contractInteraction/mainnet_contracts.json')
-    elif this_network == "testnet":
+    elif this_network == "rsk-testnet":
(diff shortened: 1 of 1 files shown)


scripts/contractInteraction/contract_interaction.py
```diff
@@ -113,21 +113,21 @@ def main():
...
+    createProposalSIP006()
+    createProposalSIP008()
 
 def loadConfig():
     global contracts, acct
     this_network = network.show_active()
     if this_network == "rsk-mainnet":
         configFile =  open('./scripts/contractInteraction/mainnet_contracts.json')
-    elif this_network == "testnet":
+    elif this_network == "rsk-testnet":
```
ororopickpocket, line 126: all good, but this line. this forces everyone to setup the "rsk-testnet".
anyway, it's kinda consistent and cleaner, so we can go through the trouble ;-)

ororopickpocket (approved):

scripts/contractInteraction/contract_interaction.py
exiledsurfer: from @powerhousefrank :

Two issuess:
1. With one address (my current one), only one proposal can be proposed at a time. So, either SIP006a or SIP008 will go through. I am thinking SIP006a is priority. (There might be another solution, but only after point 2 that would be useful).

2. SIP006a was successfully proposed in Testnet, but for Mainnet the same thing is giving trouble, I am still looking at it, but not sure will be able to do it any time soon.

TravinImmunefi:
We don't have any issue with SIP-006a going ahead of SIP-008 and if it needs to go alone first.

PR #168, opened 2021-03-09
Script for origin vesting contracts

jameshowlett977:


scripts/deployment/vesting/deploy_vestings.py
```diff
@@ -0,0 +1,107 @@
...
+            vestingList.append([tokenOwner, amount])
+
+    with open('./scripts/deployment/vesting/BTC to be returned(R&U).csv', 'r') as file:
+        reader = csv.reader(file)
+        for row in reader:
+            # tokenOwner = row[1].replace(" ", "")
+            tokenOwner = row[1]
+            amount = int(int(float(row[2]) * 10**8 * 10**18) / TOKEN_PRICE)
+            btcAmount += float(row[2])
+            vestingList.append([tokenOwner, amount])
```
ororopickpocket, line 75: just double checking: there are no wallets existing on both lists, right?

ororopickpocket (commented):
Nice. Good idea to save transactions with the OriginVestingCreator. Approved if there are for sure no duplicates on the list.

scripts/deployment/vesting/deploy_vestings.py
jameshowlett977: yes, vesting creation will fail in this case

`function createVesting(address _tokenOwner, uint256 _amount, uint256 _cliff, uint256 _duration) public onlyOwner {
        require(_tokenOwner != address(0), "Invalid address");
        require(!processedList[_tokenOwner], "Already processed");

        processedList[_tokenOwner] = true;
`

I've run it locally without any errors

PR #170, opened 2021-03-12
Governance Interaction Script Update

powerhousefrank:
Task List

- [x] Added delegatee to values.json and interaction script.
- [x] Updated the README


scripts/governance/README.md
```diff
@@ -73,9 +73,21 @@ brownie run scripts/governance/create_proposal.py --network rsk-testnet
 brownie run scripts/governance/last_proposal_created.py --network rsk-testnet
 ```
 
+## Values.json
+
+- account : This is the account address used for checking the current voting power of that account, last proposal created by an account and also the account used for staking.
+- delegatee : The address of the delegatee (who will have the right to vote for the staked token).
```
john-light, line 79: Is this referring to the address that will be able to vote on behalf of the account that I set in the `account` field? If so, I think the proper term would be "delgate" rather than "delegatee".

PR #177, opened 2021-03-25
transfer tokens to N * 100 users

jameshowlett977:
1. script (deploy_token_sender.py) and csv file (SCNC.csv) will be changed
2. TokenSender - sends tokens to N users (300+ in our case)
We need this contract to send tokens in one transaction instead of N transactions,
gas usage for 500 users: 5425397


contracts/governance/Vesting/TokenSender.sol
```diff
@@ -0,0 +1,74 @@
...
+     */
+    function transferSOV(address _receiver, uint256 _amount) public onlyAuthorized {
+        _transferSOV(_receiver, _amount);
+    }
+
+    function _transferSOV(address _receiver, uint256 _amount) internal {
+        require(_receiver != address(0), "receiver address invalid");
+        require(_amount != 0, "amount invalid");
+
+        IERC20(SOV).transfer(_receiver, _amount);
```
eMarchenko, line 70: formally, you [MUST](https://eips.ethereum.org/EIPS/eip-20#methods) check return value of ERC20 token

eMarchenko (approved):

contracts/governance/Vesting/TokenSender.sol
jameshowlett977: agree, updated, thanks

Issue #178, opened 2021-03-27
Governance Issue: `withdrawTokens` function succeed, but no transfer of tokens was performed


jjmr007:
A user did the following tx:

https://explorer.rsk.co/tx/0xb7e75a895a65b2ff50617eaec206ca0961ad74a925ab2efc69b2b9efd296fc2c?__ctab=Transaction

But no `TokensWithdrawn` event was emitted

User: `0x3B179Bd3B312050892E54d323c354f8787Fb2814`
the specific `vestingLogic` contract was: `0x9308305F7a16Edb124400BF8Fe3b4c0cAa18A095`

My main concern is verify that the user's tokens are not lost, and verify that there is no bug in these contracts
In second place: where did this user's tokens go?

jameshowlett977:
Fixed in SIP-0016

PR #180, opened 2021-03-29
Fix-Vesting(Origin).withdrawTokens

jameshowlett977:
https://github.com/DistributedCollective/Sovryn-smart-contracts/issues/178

Fix for the issue 178
===
Lock date for Origin vested tokes is 1616754495 -> GMT: Friday, March 26, 2021 10:28:15 AM
VestingLogic passes `startDate` + `cliff` as a timestamp
startDate = 1614335295 -> GMT: Friday, February 26, 2021 10:28:15 AM
cliff = 1776077
startDate + cliff = 1616111372 -> GMT: Thursday, March 18, 2021 11:49:32 PM

===
How to fix:

1. Adjust date using Staking.timestampToLockDate
2. Add 2 weeks (we need to select end of period, by default it will be start of period) if dates before and after adjustment are different (it's the case only for Origin Vesting contracts).

===
Checked using script `check_origin_vestings.py`
It's better to check on real data rather than using unit tests, it's an integration issue.
`checked-origin-list.csv` - script result


tjcloa (commented):
lgtm

eMarchenko (approved):
That's a dirty hack that should be refactored later. Considering tight schedule it is ok (sort of).
Please note that it's difficult to prove whether there are any other missed edge cases.

jameshowlett977:
> That's a dirty hack that should be refactored later. Considering tight schedule it is ok (sort of).
> Please note that it's difficult to prove whether there are any other missed edge cases.

Agree, but we don't have another option to unlock Origin Vesting contracts.
This fix shouldn't affect any vesting with correct dates.
Also it can be removed as soon as all tokens of Origin Vestings claimed.

PR #204, opened 2021-05-04
CI Update

powerhousefrank:
Currently, we are hitting the execution limit of 50 minutes in Travis, so removing the tests and using the coverage for checking the tests as well.


powerhousefrank:
Changes Introduced:
- Original Travis divided, and now only have `coverage` in it. (The current running time is at max 24 minutes)
- Github Actions is now being used for tests. (The current running time is at max 13 minutes)
- README updated to reflect the build status and test coverage.
- `hardhat-docgen` added to create technical documentation with inline coding (will perfectly fit with @computerphysicslab 's recent documentation task).
- Some prettier commands updated (write and check commands added).

powerhousefrank:
IMPORTANT: To speed up the CI, I am using `npm ci` instead of `npm i`, which makes use of package-lock.json.

Thus from now onwards, we should install all packages using `npm`, and not from `yarn`.

ororopickpocket (approved):

PR #221, opened 2021-05-07
LM rewards into lockedSOV

computerphysicslab:
The function _transferReward instead of transferring the reward to the liquidity provider (user), it deposits the reward into the lockedSOV vault contract. This implies removing SVR token from LiquidityMining contract and replace it with SOV token. Tests are to be adapted to use SOV instead of SVR.


ororopickpocket (approved):
Going to approve and merge this PR, but unit tests for `setLockedSOV` are missing. Please add them.

PR #238, opened 2021-05-18
Trading rebates using lockedsov

santosinarp:
This pr should be merge after the Affiliates_v2.
The deployment step would be:
1. Set ProtocolTokenAddress by SOV address
2. Set LockedSOVAddress -- this supposed to be done as we already set LockedSOVAddress when deploying affiliates_v2
3. Redeploy modules which implement InterestUser and SwapsUser.

Deployment script is available in contract_interaction.py -- deployTradingRebatesUsingLockedSOV()


contracts/swaps/SwapsUser.sol
```diff
@@ -171,7 +172,8 @@ contract SwapsUser is State, SwapsEvents, FeesHelper {
 				_payTradingFee(
 					addrs[4], /// user
 					loanId, /// loanId,
-					addrs[1], /// destToken
+					addrs[1], /// destToken (feeToken)
```
korepkorep, line 175: Is it correct that `specialRebates[sourceToken][destToken]` = `specialRebates[destToken][sourceToken]` since you pass `addrs[0]` as destination and `addrs[1]` as source?

korepkorep (commented):

eMarchenko (commented):
@korepkorep performed the review, there is a [possible bug](https://github.com/DistributedCollective/Sovryn-smart-contracts/pull/238#discussion_r670427246)

contracts/swaps/SwapsUser.sol
santosinarp: specialRebates[sourceToken][destToken] != specialRebates[destToken][sourceToken]
But can refer to the previous version where #1 _payTradingFee use address[0] as feeToken
https://github.com/DistributedCollective/Sovryn-smart-contracts/blob/development/contracts/swaps/SwapsUser.sol#L134

Meanwhile the #2 _payTradingFee use address[1] as feeToken
https://github.com/DistributedCollective/Sovryn-smart-contracts/blob/development/contracts/swaps/SwapsUser.sol#L174

PR #239, opened 2021-05-18
Proxy eip 1967

computerphysicslab:
Proxy contract has been modified to comply with EIP-1967 https://eips.ethereum.org/EIPS/eip-1967 , that basically standardises the storage slot for the logic contract address and for the admin account that is allowed to modify it, if upgradeable (only UpgradableProxy has an external method to do so). It requires also a renaming for the associated events.

Previous code was using slots for both address, but were not the standard ones.


ororopickpocket (commented):
We are not going to merge this PR, because we will not redeploy the proxy.

PR #255, opened 2021-06-01
Not in the same block

### LINKED ISSUE #199: Prevent lending and withdrawal within the same tx/block
To protect from the lending fees manipulation using flash loan we need to prevent lending and withdrawing from the pools in the same tx/block

computerphysicslab:
Prevent lending and withdrawal within the same tx/block #199 (medium):

To protect from the lending fees manipulation using flash loan we need to prevent lending and withdrawing from the pools in the same tx/block by the same account.

Check more details at https://github.com/DistributedCollective/Sovryn-smart-contracts/issues/199

### DIFF
--- a/contracts/connectors/loantoken/LoanTokenBase.sol
+++ b/contracts/connectors/loantoken/LoanTokenBase.sol
@@ -74,4 +74,9 @@ contract LoanTokenBase is ReentrancyGuard, Ownable, Pausable {
 	/// The maximum trading/borrowing/lending limit per token address.
 	/// 0 -> no limit
 	mapping(address => uint256) public transactionLimit;
+
+	/// @notice A mapping of accounts stores those who
+	///   lend/withdraw on current block.
+	/// block => account => true/false
+	mapping(uint256 => mapping(address => bool)) internal hasInteracted;
 }
--- a/contracts/connectors/loantoken/LoanTokenLogicStandard.sol
+++ b/contracts/connectors/loantoken/LoanTokenLogicStandard.sol
@@ -67,6 +67,8 @@ contract LoanTokenLogicStandard is LoanTokenSettingsLowerAdmin {
 	/**
 	 * @notice Mint loan token wrapper.
 	 * Adds a check before calling low level _mintToken function.
+	 * This function is called by a user to provide liquidity to a loan pool.
+	 * To get back the provided underlying tokens, the user calls burn function.
 	 * The function retrieves the tokens from the message sender, so make sure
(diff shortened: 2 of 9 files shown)


contracts/connectors/loantoken/LoanTokenBase.sol
```diff
@@ -74,4 +74,9 @@ contract LoanTokenBase is ReentrancyGuard, Ownable, Pausable {
 	/// The maximum trading/borrowing/lending limit per token address.
 	/// 0 -> no limit
 	mapping(address => uint256) public transactionLimit;
+
+	/// @notice A mapping of accounts stores those who
+	///   lend/withdraw on current block.
+	/// block => account => true/false
+	mapping(uint256 => mapping(address => bool)) internal hasInteracted;
```
eMarchenko, line 81: If I understand correctly, you can simplify it to `mapping(address => uint256) internal lastInteraction; // account => block ` 
and the check should be `lastInteraction[addr] < block.number`
it will save some gas, mostly due to storage reuse

contracts/connectors/loantoken/LoanTokenLogicStandard.sol
```diff
@@ -1382,6 +1398,27 @@ contract LoanTokenLogicStandard is LoanTokenSettingsLowerAdmin {
...
+	 * tx/block by the same account.
+	 * */
+	function _checkNotInTheSameBlock() internal {
+		uint256 _currentBlock;
+
+		/// @dev Get and buffer current block.
+		_currentBlock = block.number;
+
+		/// @dev Check there are no previous txs in this block coming from same account.
+		require(!hasInteracted[_currentBlock][msg.sender], "Avoiding flash loan attack: several txs in same block from same account.");
```
eMarchenko, line 1416: checking against `msg.sender` doesn't protect against FL, only complicates the setup:
the attacker would need to add extra contracts A1, A2, ... for each individual interaction within their transaction 

formally, you can achieve FL protection via `tx.origin` check, however it is discouraged.

eMarchenko (commented):
Flash Loans only help some attacks, but are not required for any of them. Therefore any FL-protection is flawed by design, as it doesn't secure contracts against whales or successful hackers (those who have lots of stolen assets).
These types of attacks should be dealt without assumption that FL is utilized.

Similarly, there are options to execute multi-transaction attacks atomically (see MEV/flashbots/...), therefore a sophisticated attacker can bypass `tx.origin` checks.

tjcloa:
it has been resolved in the PR#476

PR #288, opened 2021-06-25
refactored contract interaction script - split into several files

ororopickpocket:


tjcloa (approved):
contract interaction functionality is in order now )

PR #289, opened 2021-06-26
Add trader in the payTradingToAffiliate event

santosinarp:
This PR is tend to add the Trader into the PayTradingToAffiliate event, so that can be used by frontend

### DIFF
--- a/contracts/modules/Affiliates.sol
+++ b/contracts/modules/Affiliates.sol
@@ -181,6 +181,7 @@ contract Affiliates is State, AffiliatesEvents {
 		if (bonusPaymentIsSuccess) {
 			emit PayTradingFeeToAffiliate(
 				referrer,
+				tx.origin, // trader
 				token,
 				isHeld,
 				tradingFeeTokenBaseAmount,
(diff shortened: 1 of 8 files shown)


contracts/modules/Affiliates.sol
```diff
@@ -181,6 +181,7 @@ contract Affiliates is State, AffiliatesEvents {
 		if (bonusPaymentIsSuccess) {
 			emit PayTradingFeeToAffiliate(
 				referrer,
+				tx.origin, // trader
```
tjcloa, line 184: **potential issue**
this can break under certain conditions
theoretically you can borrow funds or open a position for a different wallet than the one you're sending with
this could be important e.g. for contracts which means that tx.origin != trader
another point is that initially the LoanToken function marginTrade and then FeesHelper function _payTradingFee accepts user as a parameter which is passed down the function calls
we should not break trough the interface boundaries and use inferred logic that as it can change at some point

**proposed solution**
add trader as a parameter to the functions when needed

contracts/modules/Affiliates.sol
```diff
@@ -191,6 +192,7 @@ contract Affiliates is State, AffiliatesEvents {
 		} else {
 			emit PayTradingFeeToAffiliateFail(
 				referrer,
+				tx.origin, // trader
```
tjcloa, line 195: add trader as a parameter to he function
see detales in the commend above #L184

santosinarp: Changes has been made

tjcloa (approved):

Issue #210, opened 2021-05-06
add pausable to all logic contracts


tjcloa:
**upd: reinstantiated**
needs to be done ASAP to set multisig as the pauser before ownership of the contracts transferred to governance which is ready and waiting for this issue to be resolved 
**upd: postponed, needs longer discussion**
emergency system to pause every single module/logic
we also need one big red 'button' which will pause sovryn
this needs to be thoroughly tested
**UPD from @computerphysicslab** 
"It'd be useful to be able to declare a sort-of contract-level modifier, that allows to inject code that executes on every call received by the contract, and executes based on raw msg data, signature, etc. This allows to provide a global behaviour to the contract, without needing to manually add a modifier to every external function, which is error-prone if one is forgotten."

    https://github.com/ethereum/solidity/issues/3864

It seems to me that a contract-level pausable modifier is already implemented in our codebase at:

    https://github.com/DistributedCollective/Sovryn-smart-contracts/blob/development/contracts/connectors/loantoken/Pausable.sol

and used at:

    https://github.com/DistributedCollective/Sovryn-smart-contracts/blob/development/contracts/connectors/loantoken/LoanTokenLogicStandard.sol#L57

  pausable(msg.sig)

computerphysicslab:
Actually bZx code comes with a pausable functionality not on contract level, but on function level.

computerphysicslab:
Next steps:

1.- Modify State and Proxy contract to get global protocol pausable

2.- and on module level as well.

3.- Modify bZx Pausable and Proxy contract to get pausable functions w/o using the modifier nor the explicit checker on every function.

computerphysicslab:
I have been testing the pausable contract with three levels of granurality but it is not yet clear how to deal with one of them, the protocol case. So, by now, in my local branch, loanToken can be paused and resumed with module or function granurality. But...

When trying to pause loanToken, feeSharingProxy functions (withdraw and transferToken) are still operative.

When trying to pause loanTokenV2, marginTrade function is still operative.
    The reason why is because the _checkPause() is only applied on the Proxy when delegating calls. marginTrade seems to be not using any proxy.

I think we need to reconsider the whole strategy, where to check if a function is paused or not. At proxy when delegating calls, at functions in their contracts, ... or hybrid approach, some at proxy, some others on their function implementation...

computerphysicslab:
I think I have found a good implementation for the pausable contract:

[...]

PR #311, opened 2021-07-09
Pause contracts

smbsp:
Implementation Details: https://docs.google.com/document/d/1Qe92Q_1t3Ajwvurd9_8RtV8RMGaq30hNyij9_g2QWYo/edit

Needs more testing.

closes: #210


contracts/interfaces/ISovryn.sol
```diff
@@ -111,6 +111,10 @@ contract ISovryn is
 
 	function setRebatePercent(uint256 rebatePercent) external;
 
+	function pause() external;
+
+	function unpause() external;
```
tjcloa, line 116: why not just one function toggleProtocolPause() ?

tjcloa:
here is an overall idea of implementation
as discussed primarily with @smbsp 
1. `_setTarget(this.swapExternal.selector, target); `
is replaced for 
`_setTarget(this.swapExternal.selector>>1, target);`
in all modules
that will lead to calling non-existing functions from the protocol, but this is controllable (see next)
2. add fallback functions to all modules which will serve as a hook prior to calling any module function
3. in the fallback function 
- check that the protocol is not on pause and virtually any preliminary processing can be done there
- unshift func sig to get the real one: msg.sig<<1 
- check that the unshifted (msg.sig<<1) sig exists
- call the func with unshifted (msg.sig<<1) signature

contracts/interfaces/ISovryn.sol
smbsp: Removed Pause/Unpause and created a new function toggleFreeze(). Added test cases for one more module - LoanSettings

korepkorep (commented):

contracts/modules/ProtocolSettings.sol
```diff
@@ -593,4 +595,9 @@ contract ProtocolSettings is State, ProtocolTokenUser, ProtocolSettingsEvents {
 	function getLockedSOVAddress() external view returns (address) {
 		return lockedSOVAddress;
 	}
+
+	function togglePaused(bool paused) external onlyOwner {
+		pause = paused;
```
denis-abag, line 600: require the pause value is changing

contracts/modules/ProtocolSettings.sol
```diff
@@ -593,4 +595,9 @@ contract ProtocolSettings is State, ProtocolTokenUser, ProtocolSettingsEvents {
...
+		emit TogglePaused(msg.sender, !paused, paused);
```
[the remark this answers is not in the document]
smbsp: Added a check to check if the parameter differs

tjcloa (approved):

PR #331, opened 2021-07-19
add harhdhat-abi-exporter

tjcloa:
updates contracts ABIs on comiple in ./artifacts/abi dir

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,4 @@ tmp/
 tmp
 /temp/
 /out/
+abi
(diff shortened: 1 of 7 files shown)


powerhousefrank (approved):
Changes:
- `path` was changed because else it would interfere with normal works, resulting in namespace collisions.
- `clear` was marked true so in the future there will not be namespace collision due to multiple abi instances.
- `spacing` was made as 4 to adhere to tab spacing formatting style.

Issue #297, opened 2021-06-30
Emit an event on toggle function pause


tjcloa:

computerphysicslab:
Sovryn-smart-contracts/contracts/connectors/loantoken/LoanTokenSettingsLowerAdmin.sol
  175,9: 	function toggleFunctionPause(

Sovryn-smart-contracts/contracts/mockup/previousLoanToken/PreviousLoanTokenSettingsLowerAdmin.sol
  114,9: 	function toggleFunctionPause(

computerphysicslab:
Adding test to check event is really emitted.

Sovryn-smart-contracts/tests-js/loan-token/Administration.test.js
  127,25: 			await localLoanToken.toggleFunctionPause(functionSignature, true);
  137,25: 			await localLoanToken.toggleFunctionPause(functionSignature, false);
  147,38: 			await expectRevert(localLoanToken.toggleFunctionPause("mint(address,uint256)", true, { from: accounts[1] }), "onlyPauser");

computerphysicslab:
Hardhat testing what events were emitted with what arguments:

According to https://ethereum-waffle.readthedocs.io/en/latest/matchers.html?highlight=events#emitting-events this code should work:

```
await expect(localLoanToken.toggleFunctionPause(functionSignature, true))
    .to.emit(localLoanToken, 'ToggleFunctionPause')
    .withArgs(functionSignature, true);
```

But it is not! So I had to use this one:

```
let tx = await localLoanToken.toggleFunctionPause(functionSignature, true);
// Check event is properly emitted
await expectEvent.inTransaction(tx.receipt.rawLogs[0].transactionHash, LoanTokenLogicStandard, "ToggleFunctionPause", {
    funcId: functionSignature,
    isPaused: true,
});
```

computerphysicslab:
After adding event, tests are not passing:
```
Error: Transaction reverted: trying to deploy a contract whose code is too large
```

computerphysicslab:
Added event ToggleFunctionPause:
https://github.com/DistributedCollective/Sovryn-smart-contracts/commit/6dec0cf13677e78a6e8a90b4a211e1cdc0c4b1b8

Added event check on test:
https://github.com/DistributedCollective/Sovryn-smart-contracts/commit/b7bcb0568ca6235c60c496374a2920727d48042b

PR #317, opened 2021-07-13
Event toggle function pause

smbsp:
closes: #297 
Added an event ToggleFunctionPaused to the function toggleFunctionPause()
Shortened some error messages to downsize the contract `LoanTokenLogicStandard`

PR #357, opened 2021-08-18
Event toggle function pause new

smbsp:
closes: #297
Added an event ToggleFunctionPaused to the function toggleFunctionPause()

We earlier had Contract Size issues and hence this version is built using updated Loan Token logic using Beacon Chain. This PR is a replacement of: https://github.com/DistributedCollective/Sovryn-smart-contracts/pull/317

### DIFF
--- a/contracts/connectors/loantoken/modules/LoanTokenSettingsLowerAdmin.sol
+++ b/contracts/connectors/loantoken/modules/LoanTokenSettingsLowerAdmin.sol
@@ -22,6 +22,7 @@ contract LoanTokenSettingsLowerAdmin is LoanTokenLogicStorage {
 	/* Events */
 
 	event SetTransactionLimits(address[] addresses, uint256[] limits);
+	event ToggleFunctionPaused(string functionId, bool prevFlag, bool newFlag);
 
 	/* Functions */
 
@@ -184,6 +185,7 @@ contract LoanTokenSettingsLowerAdmin is LoanTokenLogicStorage {
 		string memory funcId, /// example: "mint(uint256,uint256)"
 		bool isPaused
 	) public {
+		bool paused;
 		require(msg.sender == pauser, "onlyPauser");
 		/// keccak256("iToken_FunctionPause")
 		bytes32 slot =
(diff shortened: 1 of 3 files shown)


contracts/connectors/loantoken/modules/LoanTokenSettingsLowerAdmin.sol
```diff
@@ -193,9 +195,14 @@ contract LoanTokenSettingsLowerAdmin is LoanTokenLogicStorage {
 					uint256(0xd46a704bc285dbd6ff5ad3863506260b1df02812f4f857c8cc852317a6ac64f2)
 				)
 			);
+		assembly {
+			paused := sload(slot)
+		}
+		require(paused != isPaused, "invalid");
```
ororopickpocket, line 201: the require is not strictly required here, but doesn't hurt either. however, the error message doesn't provide any valuable information. change it to "isPaused is already set to that value" or similar message.

contracts/connectors/loantoken/modules/LoanTokenSettingsLowerAdmin.sol
```diff
@@ -193,9 +195,14 @@ contract LoanTokenSettingsLowerAdmin is LoanTokenLogicStorage {
...
 		assembly {
 			sstore(slot, isPaused)
 		}
+		emit ToggleFunctionPaused(funcId, !isPaused, isPaused);
```
ororopickpocket, line 205: the event should rather be called
ToggledFunctionPause
(toggle is the verb here)

ororopickpocket (changes requested):

contracts/connectors/loantoken/modules/LoanTokenSettingsLowerAdmin.sol
smbsp: Done

contracts/connectors/loantoken/modules/LoanTokenSettingsLowerAdmin.sol
smbsp: Done

ororopickpocket (approved):

tjcloa (approved):

Alitasovryn (commented):
Looks fine.

Issue #354, opened 2021-08-04
Add a isContract check


smbsp:
_callOptionalReturn function doesn't have a check to verify if it is a Contract Account or Externally Owned Account.

PR #358, opened 2021-08-18
Add is contract check new

smbsp:
closes: https://github.com/DistributedCollective/Sovryn-smart-contracts/issues/354

Added a check that the token address passed must be a contract. At the moment when we pass EOA address, the success will return true and returndata.length = 0, which bypass the code. While this is not an immediate threat, it is good to have this check in the internal _callOptionalReturn() function.

The changes were pending due to contract size issue.

### DIFF
--- a/contracts/connectors/loantoken/modules/LoanTokenLogicStandard.sol
+++ b/contracts/connectors/loantoken/modules/LoanTokenLogicStandard.sol
@@ -1257,7 +1257,7 @@ contract LoanTokenLogicStandard is LoanTokenLogicStorage {
 		bytes memory data,
 		string memory errorMsg
 	) internal {
-		require(Address.isContract(token), "call to non-contract");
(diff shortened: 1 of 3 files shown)


contracts/connectors/loantoken/modules/LoanTokenLogicStandard.sol
```diff
@@ -1257,6 +1257,7 @@ contract LoanTokenLogicStandard is LoanTokenLogicStorage {
 		bytes memory data,
 		string memory errorMsg
 	) internal {
+		require(Address.isContract(token), "call to non-contract");
```
tjcloa, line 1260:
```suggestion
		require(Address.isContract(token), "call to a non-contract address");
```

ororopickpocket (approved):

tjcloa (approved):

Alitasovryn (commented):
Looks fine.

PR #361, opened 2021-08-27
Reverse funds iwrbtc

smbsp:
Task: https://taiga.sovryn.app/project/sovryn-order/us/4?kanban-status=36

Owner can now withdraw all rBTC stuck in Loan Token contracts using a new function withdrawAllRBTC().

Background: When someone is interacting with the loan token contracts using a wrong ABI sending RBTC, the RBTC can get stuck on the contract. When used as intended, the loan token contracts never hold any RBTC, because all RBTC are immediately wrapped to RBTC. So, if the contract is holding RBTC, they ended up there by mistake and should hence be recoverable.

### DIFF
--- a/contracts/connectors/loantoken/modules/LoanTokenLogicStandard.sol
+++ b/contracts/connectors/loantoken/modules/LoanTokenLogicStandard.sol
@@ -430,10 +430,10 @@ contract LoanTokenLogicStandard is LoanTokenLogicStorage {
 	 * @param _amount The amount of rBTC to be transferred.
 	 */
 	function withdrawRBTCTo(address payable _receiverAddress, uint256 _amount) external onlyOwner {
-        require(_receiverAddress != address(0), "receiver address invalid");
-        require(_amount > 0, "non-zero withdraw amount expected");    
-        require(_amount <= address(this).balance, "withdraw amount cannot exceed balance");
-        _receiverAddress.transfer(_amount);
+		require(_receiverAddress != address(0), "receiver address invalid");
+		require(_amount > 0, "non-zero withdraw amount expected");
+		require(_amount <= address(this).balance, "withdraw amount cannot exceed balance");
+		_receiverAddress.transfer(_amount);
 	}
 
 	/**
--- a/contracts/connectors/loantoken/modules/LoanTokenLogicStandard.sol
+++ b/contracts/connectors/loantoken/modules/LoanTokenLogicStandard.sol
@@ -50,6 +50,10 @@ contract LoanTokenLogicStandard is LoanTokenLogicStorage {
 	using SafeMath for uint256;
 	using SignedSafeMath for int256;
 
+	/* Events */
+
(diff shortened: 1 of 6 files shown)


tjcloa (changes requested):
what if we have more than one amount from different addresses?
pls refactor: 
1. add function `withdrawRBTC(address, amount)`
2. call the function `withdrawRBTC` from `withdrawAllRBTC` setting amount = lending pool balance

smbsp:
> what if we have more than one amount from different addresses?
> pls refactor:
> 
> 1. add function `withdrawRBTC(address, amount)`
> 2. call the function `withdrawRBTC` from `withdrawAllRBTC` setting amount = lending pool balance

Modified the code to accommodate transfer of different amounts to multiple addresses

contracts/connectors/loantoken/modules/LoanTokenLogicStandard.sol
```diff
@@ -417,6 +417,18 @@ contract LoanTokenLogicStandard is LoanTokenLogicStorage {
...
+	/**
+	 * @notice Withdraws RBTC from the contract by Multisig.
+	 * @param _receiverAddress The address where the rBTC has to be transferred.
+	 * @param _amount The amount of rBTC to be transferred.
+	 */
+	function withdrawRBTCTo(address payable _receiverAddress, uint256 _amount) external onlyOwner {
+		require(_receiverAddress != address(0), "receiver address invalid");
+		require(_amount > 0, "non-zero withdraw amount expected");
+		require(_amount <= address(this).balance, "withdraw amount cannot exceed balance");
+		_receiverAddress.transfer(_amount);
```
eMarchenko, line 429: consider emitting an event to simplify monitoring.

contracts/connectors/loantoken/modules/LoanTokenLogicWrbtc.sol
```diff
@@ -22,7 +22,7 @@ contract LoanTokenLogicWrbtc is LoanTokenLogicStandard {
 	 * @return The list of function signatures (bytes4[])
 	 */
 	function getListFunctionSignatures() external pure returns (bytes4[] memory functionSignatures, bytes32 moduleName) {
-		bytes4[] memory res = new bytes4[](35);
+		bytes4[] memory res = new bytes4[](36);
```
eMarchenko, line 25: `withdrawRBTCTo` probably should never be called for wrbtc loans. If it is true, consider moving the function into LoanTokenLogicLM.

eMarchenko (commented):
I'm not sure how one can mistakenly send rbtc into wrong Loan.
Anyway, the implementation of withdraw function is good.

contracts/connectors/loantoken/modules/LoanTokenLogicStandard.sol
smbsp: Done

tjcloa (approved):
lgtm

Alitasovryn (commented):
Looks good.

PR #364, opened 2021-09-06
Separate swap external fee percentage from using trading fee percentage

cwsnt:


ororopickpocket (approved):

contracts/core/State.sol
```diff
@@ -204,6 +204,8 @@ contract State is Objects, ReentrancyGuard, Ownable {
 	mapping(address => mapping(address => uint256)) public specialRebates; // Special rate rebates for spesific pair -- if not set, then use the default one
 	bool public pause; //Flag to pause all protocol modules
 
+	uint256 internal swapExtrernalFeePercent;
```
eMarchenko, line 207: I guess you must add comments to all new storage variables and functions

contracts/modules/ProtocolSettings.sol
```diff
@@ -234,6 +236,19 @@ contract ProtocolSettings is State, ProtocolTokenUser, ProtocolSettingsEvents, M
 		emit SetBorrowingFeePercent(msg.sender, oldValue, newValue);
 	}
 
+	/**
+	 * @notice Set the value of swapExtrernalFeePercent storage variable
+	 *
+	 * @param newValue the new value for swapExternalFeePercent
+	 */
+	function setSwapExternalFeePercent(uint256 newValue) external onlyOwner whenNotPaused {
```
eMarchenko, line 244: consider giving the `newValue` parameter a more meaningful name.
// there are multiple setters with  `newValue` parameter

contracts/swaps/SwapsUser.sol
```diff
@@ -99,7 +100,8 @@ contract SwapsUser is State, SwapsEvents, FeesHelper {
 		uint256[3] memory vals,
```
eMarchenko, line 100: consider using structs instead of arrays to reduce argument number. 
iirc it should work at least for internal functions (I'm not sure if 0.5.17 compiler allows it)

contracts/swaps/SwapsUser.sol
```diff
@@ -99,7 +100,8 @@ contract SwapsUser is State, SwapsEvents, FeesHelper {
...
 		bytes32 loanId,
 		bool miscBool, /// bypassFee
-		bytes memory loanDataBytes
+		bytes memory loanDataBytes,
+		bool isSwapExternal
```
eMarchenko, line 104: adding extra flag isn't elegant. this functions slowly grows into if-else hell, consider refactoring it in the future

eMarchenko (commented):
The update is good, but some parts of code start to degrade

PR #365, opened 2021-09-06
Change trading rebate rewards SOV become liquid

cwsnt:
Enable setter & getter for locked sov basis point
Added Deployment script & unit test for the setter & getter


contracts/core/State.sol
```diff
@@ -204,6 +204,8 @@ contract State is Objects, ReentrancyGuard, Ownable {
 	mapping(address => mapping(address => uint256)) public specialRebates; // Special rate rebates for spesific pair -- if not set, then use the default one
 	bool public pause; //Flag to pause all protocol modules
 
+	uint256 internal tradingRebateRewardsBasisPoint;
```
ororopickpocket, line 207: all variables which are added should be documented.
e.g.: 
Defines the portion of the trading fees which is to be paid out in a liquid form in basis points (0.01%). The rest is vested.

contracts/modules/ProtocolSettings.sol
```diff
@@ -114,6 +116,15 @@ contract ProtocolSettings is State, ProtocolTokenUser, ProtocolSettingsEvents, M
 		emit SetLockedSOVAddress(msg.sender, oldLockedSOVAddress, newLockedSOVAddress);
 	}
 
+	function setTradingRebateRewardsBasisPoint(uint256 newBasisPoint) external onlyOwner whenNotPaused {
```
ororopickpocket, line 119: documentation missing.
never add functions without documentation.

contracts/modules/ProtocolSettings.sol
```diff
@@ -651,4 +662,8 @@ contract ProtocolSettings is State, ProtocolTokenUser, ProtocolSettingsEvents, M
 	function isProtocolPaused() external view returns (bool) {
 		return pause;
 	}
+
+	function getTradingRebateRewardsBasisPoint() external view returns (uint256) {
```
ororopickpocket, line 666: same here

contracts/modules/ProtocolSettings.sol
```diff
@@ -114,6 +116,15 @@ contract ProtocolSettings is State, ProtocolTokenUser, ProtocolSettingsEvents, M
 		emit SetLockedSOVAddress(msg.sender, oldLockedSOVAddress, newLockedSOVAddress);
 	}
 
+	function setTradingRebateRewardsBasisPoint(uint256 newBasisPoint) external onlyOwner whenNotPaused {
+		require(newBasisPoint <= 10000, "value too high");
```
ororopickpocket, line 120: lockedSOV only allows 9999, does it not? So, we should have the same restriction here.

ororopickpocket (changes requested):

eMarchenko (approved):

PR #367, opened 2021-09-08
use min initial margin from loan params for borrowing

ororopickpocket:
previously the loan params settings were overwritten with a hardcoded margin for borrowing.
now, it's using the settings.

because of code size issues, I removed the functions for setting transaction limits and changing the token name and symbol from the LoanTokenSettingsLowerAdmin. These functions may be reintroduced after refactoring.


contracts/connectors/loantoken/LoanTokenLogicStandard.sol
```diff
@@ -370,11 +370,13 @@ contract LoanTokenLogicStandard is LoanTokenSettingsLowerAdmin {
 			sentAmounts[1] /// depositAmount
 		);
 
+		/// @dev Converting to initialMargin
+		leverageAmount = SafeMath.div(10**38, leverageAmount);
```
eMarchenko, line 374: this like doesn't look great - you should declare a new variable instead of reusing existing one.
I guess you are trying to avoid "Stack too deep" issue here, consider inlining this logic into return statement below

eMarchenko (commented):
I'm just hoping that you are not disabling tests to speed up the process...

contracts/connectors/loantoken/LoanTokenLogicStandard.sol
ororopickpocket: tried that too. also gave a stack too deep error.

ororopickpocket: btw: only commented out the tests for the removed functions - which could be reintroduced with the refactored version (code size issues).

PR #368, opened 2021-09-08
Fix swaps error when sending rbtc

cwsnt:
Fix internalSwap error when sending RBTC
Cover unit testing for using rbtc to do the swap


ororopickpocket (approved):

contracts/swaps/connectors/testnet/SwapsImplLocal.sol
```diff
@@ -39,7 +39,7 @@ contract SwapsImplLocal is State, ISwapsImpl {
 		uint256 minSourceTokenAmount,
 		uint256 maxSourceTokenAmount,
 		uint256 requiredDestTokenAmount
-	) public returns (uint256 destTokenAmountReceived, uint256 sourceTokenAmountUsed) {
+	) public payable returns (uint256 destTokenAmountReceived, uint256 sourceTokenAmountUsed) {
```
eMarchenko, line 42: I don't see where RBTC leaves the contract. `msg.value` or `balance` are never used, neither are `.transfer`, `.send`, or `.value`

eMarchenko (commented):
I don't know how it can work with RBTC (I might be just be wrong)

contracts/swaps/connectors/testnet/SwapsImplLocal.sol
ororopickpocket: this contract is called using delegateCall from the protocol. the first function called is on the SwapsExternal contract. It receives RBTC, wraps them to WRBTC and then wants to swap the WRBTC. so, msg.value is used on the delegate caller. without the payable modifier the transaction fails on the delegatecall. i believe that's because the delegate called contract uses the msg.value from the caller.

eMarchenko: oh, I see it. thank you for the detailed explanation.
So `A.foo` delegatecalls `B.bar`, and `mag.value` is preserved. If `B.bar` is non-payable, then the whole transaction fails (Solidity compiler inserts `msg.value == 0` check for non-payable functions).

PR #370, opened 2021-09-10
Sov collateral for borrowing

ororopickpocket:
I needed to change the `PriceFeedV1PoolOracle`, because it was outdated and no longer compatible with v1 pool oracles from the oracle-based-amm repository.

The old  `PriceFeedV1PoolOracle` assumed `latestAnswer` to return the price with WRBTC as quote currency, but instead it is used as base currency. Therefore `latestPrice` is used instead which allows us to pass the base currency. Also, the precision changed. The price is now returned with 18 decimals.


eMarchenko (commented):
looks good

PR #377, opened 2021-09-18
Multiple rewards Liquidity Mining

swamp-thing-sovryn:
# Multiple rewards Liquidity Mining

## Goal

A new liquidity mining contract will be deployed that will let users to get multiple tokens in exchange of staking LP Tokens as opposed to the current version which only rewards `SOV`:

1. Different reward tokens can be configured by the contract administrator. Reward amiunt per block can be configured depending when doing so.
2. LP Pools can be configured with different allocation points, i.e. weight, and be associated to multiple reward tokens.
3. Users, by depositing LP tokens, will get as many different tokens based on the aforementioned quantities.

## Implementation

Upgrading the original `LiquidityMining.sol` contract would have been extremely complex the storage layout would require lots of changes. In order to make users experience smoother and not to make them to actively move their funds to the new contract and pay for the migration trnasaction fees, the following plan is proposed:

1. Implement the new liquidity mining contract as a brand new deployment, named `LiquidityMiningV2`.
2. Replace `LiquidityMining` contract with a new version that contains a new set of capabilities:
  a. Allows admins to stop some operations (see `3`)
  b. Exposes a new interface to extract some data from it's state
3. Stop rewarding users by invoking `stopMining`
4. Define a period where users can opt-out of this automatic migration. Deposits will be forbidden but the users will be able to claim the rewards or even withdraw all their funds if they consider so.
5. One that period has finished, the `LiquidityMining.sol` contract will be locked the users will not be able to neither withdraw or claim rewards.
6. The admin will invoke a function on `LiquidityMiningV2` that will:
  a. Copy the original pools and rewards configurations into the new liquidity mining contract
  b. Move the remaining `SOV` and `LP Tokens` stored in `LiquidityMiningV1`
7. Update the GUI to let the users interact with the new contract

sequence diagram

state diagram

## Deployment steps

1. Replace `LiquidityMining` implementation by `LiquidityMiningV1`
2. Deploy `LiquidityMiningV2`
3. Add `LMV1toLMV2Migrator` as `LiquidityMiningV2` and `LiquidityMiningV1` admin
4. Disable `LiquidityMiningV1` deposits
5. Let the users migrate and then invoke the migration functions to copy data from the old contract and transfer funds

### DIFF
--- a/contracts/farm/LiquidityMiningV1.sol
+++ b/contracts/farm/LiquidityMiningV1.sol
@@ -36,6 +36,14 @@ contract LiquidityMiningV1 is ILiquidityMiningV1, LiquidityMiningStorageV1 {
 		_;
 	}
 
+	modifier onlyMigrationGracePeriodStarted() {
+		require(
+			migrationGracePeriodState == MigrationGracePeriodStates.Started,
+			"Forbidden: Migration hasn't started yet or already finished"
+		);
+		_;
+	}
+
 	modifier onlyBeforeMigrationGracePeriodFinished() {
 		require(migrationGracePeriodState < MigrationGracePeriodStates.Finished, "Forbidden: contract deprecated");
 		_;
@@ -56,6 +64,7 @@ contract LiquidityMiningV1 is ILiquidityMiningV1, LiquidityMiningStorageV1 {
 	function initialize(address _liquidityMiningV2) external onlyAuthorized {
 		/// @dev Non-idempotent function. Must be called just once.
 		require(_liquidityMiningV2 != address(0), "Invalid address");
+		require(liquidityMiningV2 == address(0), "Already initialized");
 		liquidityMiningV2 = _liquidityMiningV2;
 	}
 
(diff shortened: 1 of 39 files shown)


contracts/farm/LiquidityMiningV1.sol
```diff
@@ -0,0 +1,767 @@
...
+	function startMigrationGracePeriod() external onlyAuthorized onlyBeforeMigrationGracePeriod {
+		migrationGracePeriodState = MigrationGracePeriodStates.Started;
+		stopMining();
+	}
+
+	// TODO: this should only be used by the LiquidityMiningV2 contract??
+	/// @notice This function finishes the migration process disabling further withdrawals and claims
+	/// @dev migration grace period should have started before this function is called.
+	function finishMigrationGracePeriod() external onlyAuthorized onlyBeforeMigrationGracePeriodFinished {
+		require(migrationGracePeriodState == MigrationGracePeriodStates.Started, "Migration hasn't started yet");
```
swamp-thing-sovryn, line 111: The `require` introduces a harder contraint making the `onlyBeforeMigrationGracePeriodFinished` worthless ... as suggested by Haku we might create a new modifier `onlyMigrationGracePerioidStarted` that checks the same condition as the `require`

py-ro: 👍

tjcloa:
this could be useful in the future, hence leaving it open

PR #387, opened 2021-10-26
Fix potential SOV clash between trading rebate reward & fees held

cwsnt:
To avoid the clash between SOV for paying the trading rebates and SOV that is used as fees (trading, borrowing, lending) in the protocol, we will introduce an additional check in the Trading Rebate functionality to only use the SOV outside the fees.

Also we will introduce a new getter function which will return the value of SOV that is used to pay the trading rebates. 
the formula will be = SOV Balance of the protocol - (trading, lending, borrowing fees)


eMarchenko (approved):

tjcloa (approved):
lgtm

PR #389, opened 2021-11-02
Update sip0024 excluding staking

smbsp:
The Staking Rewards logic have the following issues:

1. The 'average block time' is used to calculate the block number of a Staking checkpoint. This might significantly change in the future and that could cause issues.
**Solution:** We are using a state variable to store 'average block time' and can update it in such scenarios
2. When we are calculating block number for a checkpoint that is far from the start time, the calculation can fluctuate and account for inconsistent rewards calculations.
**Solution:** We use a mapping to store block number of a checkpoint and the calculation is triggered immediately after the checkpoint is passed(potentially using a CRON job) for more accurate block number. This can be triggered by anyone. If the block number calculation is not triggered, the rewards calculations will still be processed as per the old logic. We also have an option to calculate block numbers for historical checkpoints to reduce the calculation/anomalies.

Also fixed inheritance order issue, improved readability and did some gas optimisations in this version.


contracts/governance/StakingRewards/StakingRewards.sol
```diff
@@ -72,14 +59,59 @@ contract StakingRewards is StakingRewardsStorage, Initializable {
...
+	}
+
+	/**
+	 * @notice This function computes the block number using the average block time for a given historical
+	 * checkpoint which is added to the mapping `checkpointBlockDetails`.
+	 * @param _time Exact staking checkpoint time
+	 */
+	function setHistoricalBlock(uint256 _time) external onlyOwner {
+		_setBlock(_time);
+	}
```
eMarchenko, line 101: if for some reason time between blocks will change several times (e.g. Ethereum's difficulty bomb), this trick might not work well, especially for older blocks. so this contract requires regular maintenance.

eMarchenko (approved):

PR #403, opened 2021-12-15
Fixing math error

smbsp:
The calculation where base rate starts scaling to max(kink level) was a bit exaggerated due to a maths errors. It has been fixed with this PR.


tjcloa (approved):
lgtm

PR #386, opened 2021-10-21
Feat/rewards cron

Genysys:

PR #405, opened 2021-12-20
Update sip0024 excluding staking cron

smbsp:
We need to run a cron job every alternate Friday after a new Staking Interval is started.

Example: A new staking interval started at 10:28:15 UTC on December 17, 2021. The next intervals starts at 
- 10:28:15 UTC on December 31, 2021
- 10:28:15 UTC on January 14, 2022
- 10:28:15 UTC on January 28, 2022 and so on.

So this cron job should subsequently run at 11 AM UTC for every interval. The cron job basically calls a function 'setBlock' that saves the checkpoint details which is useful for calculating the Staking Rewards.

The changes are done with help from: https://github.com/DistributedCollective/Sovryn-smart-contracts/pull/386


scripts/contractInteraction/config.py
```diff
@@ -21,6 +21,19 @@ def loadConfig():
...
+        print("Running cron job")
+        acct = accounts.add(
+            environ.get('FEE_CLAIMER'))
+        configFile = open(
+            './scripts/contractInteraction/testnet_contracts.json')
+    elif thisNetwork == "rsk-mainnet" and environ.get('REWARDS_CRON') == "1":
+        acct = accounts.add(
+            environ.get('FEE_CLAIMER'))
+        configFile = open(
+            './scripts/contractInteraction/testnet_contracts.json')
```
ororopickpocket, line 34: shouldn't this read mainnet_contracts?

smbsp: Resolved

PR #407, opened 2022-01-06
Fix double received feesharing

cwsnt:
- Fix WRBTC double spending when withdrawing fees from protocol to the FeeSharingProxy which causes wrbtc stuck in the FeeSharingProxy.
- Adding wrbtc withdrawal function in FeeSharingProxy to withdraw the stuck wrbtc

### DIFF
--- a/contracts/modules/ProtocolSettings.sol
+++ b/contracts/modules/ProtocolSettings.sol
@@ -409,7 +409,7 @@ contract ProtocolSettings is State, ProtocolTokenUser, ProtocolSettingsEvents, M
 					(amountConvertedToWRBTC, ) = ProtocolSwapExternalInterface(protocolAddress).swapExternal(
 						tokens[i], // source token address
 						address(wrbtcToken), // dest token address
-						feesController, // set feeSharingProxy as receiver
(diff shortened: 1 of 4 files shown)


contracts/modules/ProtocolSettings.sol
```diff
@@ -403,13 +403,15 @@ contract ProtocolSettings is State, ProtocolTokenUser, ProtocolSettingsEvents, M
...
+
+					IERC20(address(wrbtcToken)).safeTransfer(receiver, amountConvertedToWRBTC);
 				} else {
 					IERC20(tokens[i]).approve(protocolAddress, tempAmount);
 
 					(amountConvertedToWRBTC, ) = ProtocolSwapExternalInterface(protocolAddress).swapExternal(
 						tokens[i], // source token address
 						address(wrbtcToken), // dest token address
-						feesController, // set feeSharingProxy as receiver
+						feesController, // set protocol as receiver
```
ororopickpocket, line 414: Receiver is still rge fee sharing proxy, isn't it? Why did you chenge the comment?

Apart from that all looks good to me.

eMarchenko (commented):
l;ooks good

PR #412, opened 2022-01-21
Feature: Create V2 struct of LoanReturnData

cwsnt:
We need to keep the old LoanReturnData struct as it is, to support the backward compatibility (especially for watchers).
At the same time, returning the borrower address & creationTimestamp in the LoanData struct can improve the scalability for the backend process.
So the solution is introducing the new struct of LoanReturnData called LoanReturnDataV2, which will contain all of the old LoanReturnData + borrower address + creationTime of the loan.

### DIFF
--- a/tests/loan-token/tradingFunctions.js
+++ b/tests/loan-token/tradingFunctions.js
@@ -94,7 +94,8 @@ const margin_trading_sending_loan_tokens = async (accounts, sovryn, loanToken, u
 	//TODO: problem: rounding error somewhere
 
 	const loan_id = args["loanId"];
-	const loan = await sovryn.getLoanV2(loan_id);
+	const loan = await sovryn.getLoan(loan_id);
+	const loanV2 = await sovryn.getLoanV2(loan_id);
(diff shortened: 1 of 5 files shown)


tests/loan-token/tradingFunctions.js
```diff
@@ -94,7 +94,7 @@ const margin_trading_sending_loan_tokens = async (accounts, sovryn, loanToken, u
 	//TODO: problem: rounding error somewhere
 
 	const loan_id = args["loanId"];
-	const loan = await sovryn.getLoan(loan_id);
+	const loan = await sovryn.getLoanV2(loan_id);
```
tjcloa, line 97: We still have `getLoan` function, so the test should be in place, please restore it

tjcloa (approved):
lgtm

ororopickpocket (approved):

Alitasovryn:
Reviewed the test cases and tests are passing. But I am not sure why the test case " Test reduce loan_duration 0 withdraw should fail" is repeated twice. Test case and test script are being repeated.

PR #416, opened 2022-02-10
Fixing sip0024 duration

smbsp:
When a new user stakes after max duration is elapsed(30 weeks) from the start time of the staking rewards program, the rewards computation doesn't work. Hence, we introduce a `restartTime` to restart the calculation from the end of the last period. This ensures that the rewards are calculated and distributed correctly.


Alitasovryn:
Reviewed the test cases and test are passing. Performed UI testing on the testnet and the rewards are accumulating as expected.

Alitasovryn (approved):

PR #418, opened 2022-03-01
Upgradable vesting

smbsp:
1. Making the four year vesting upgradable.
2. Enabling token owner change

Both the items should be approved by both vesting owner as well as token owner.


contracts/governance/Vesting/fouryear/FourYearVestingLogic.sol
```diff
@@ -137,6 +137,41 @@ contract FourYearVestingLogic is IFourYearVesting, FourYearVestingStorage, Appro
...
+
+	/**
+	 * @notice Change token owner.
+	 * @dev Changes token owner. Must be
+	 * approved by both token owner and multisig(vesting owner).
+	 * @param _newTokenOwner Address of new token owner.
+	 * */
+	function changeTokenOwner(address _newTokenOwner) public onlyOwner {
+		require(_newTokenOwner != address(0), "invalid token owner address");
+		require(signed[tokenOwner], "must be signed by token owner");
```
ororopickpocket, line 159: here it is required that the tokenowner approves first before the change was suggested. 

it should be the other way round. First, a new owner is suggested, then confirmed.

```
function changeTokenOwner(address _newTokenOwner) public onlyOwner{ 
    //contract owner suggests the ownership transfer, add eequirement checks
    newOwner = _newTokenOwner;
}

function approveOwnershipTransfer() public onlyTokenOwner{
     require(newTokenOwner != address(0));
     tokenOwner = newTokenOwner;
}
```

ororopickpocket (changes requested):

contracts/governance/Vesting/fouryear/FourYearVestingLogic.sol
smbsp: Done

contracts/governance/Vesting/fouryear/FourYearVestingLogic.sol
```diff
@@ -137,6 +137,41 @@ contract FourYearVestingLogic is IFourYearVesting, FourYearVestingStorage, Appro
 		emit DividendsCollected(msg.sender, _loanPoolToken, _receiver, _maxCheckpoints);
 	}
 
+	/**
+	 * @notice Sign transactions - only Token Owner.
+	 * @dev The setImpl and changeTokenOwner functions can only be
+	 * executed when both vesting owner and token owner have approved. This
+	 * function ascertains the approval of token owner.
+	 * */
+	function signTransaction() public onlyTokenOwner {
```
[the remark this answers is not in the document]
smbsp: Modified to change(token owner) -> approve(vesting owner) pattern

contracts/governance/Vesting/fouryear/FourYearVesting.sol
```diff
@@ -54,10 +55,25 @@ contract FourYearVesting is FourYearVestingStorage, Proxy {
...
-	function governanceWithdrawTokens(address receiver) public {
+	function governanceWithdrawTokens(address receiver) public pure {
 		revert("operation not supported");
 	}
+
+	/**
+	 * @notice Set address of the implementation.
+	 * @dev Overriding setImpl function of implementation. The logic can only be
+	 * modified when both token owner and veting owner approve. Since
+	 * setImplementation can only be called by vesting owner, we only need to check
```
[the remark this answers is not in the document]
smbsp: Modified to change(token owner) -> approve(vesting owner) pattern

contracts/governance/Vesting/fouryear/FourYearVesting.sol
```diff
@@ -55,10 +56,26 @@ contract FourYearVesting is FourYearVestingStorage, Proxy {
...
+	/**
+	 * @notice Set address of the implementation - vesting owner.
+	 * @dev Overriding setImplementation function of UpgradableProxy. The logic can only be
+	 * modified when both token owner and veting owner approve. Since
+	 * setImplementation can only be called by vesting owner, we also need to check
+	 * if the new logic is already approved by the token owner.
+	 * @param _implementation Address of the implementation. Must match with what is set by token owner.
+	 * */
+	function setImplementation(address _implementation) public onlyProxyOwner {
+		require(newImplementation != address(0), "invalid address");
```
tjcloa, line 76: pls use oz Address library to check if the address is a contract or use EIP-165 to verify if the contract supports expected interfaces

contracts/governance/Vesting/fouryear/FourYearVesting.sol
```diff
@@ -55,10 +56,26 @@ contract FourYearVesting is FourYearVestingStorage, Proxy {
 	}
```
tjcloa, line 57: L42-43 above
```
require(_cliff == 4 weeks, "invalid cliff");
require(_duration == 156 weeks, "invalid duration");
```
these are constants semantically - no  need to consume storage space
and addresses verification - pls use OZ Address lib to check if the address is a contract or verify required interface by using EIP-165

contracts/governance/Vesting/fouryear/FourYearVestingFactory.sol
```diff
@@ -26,6 +26,7 @@ contract FourYearVestingFactory is IFourYearVestingFactory, Ownable {
 	 * @param _duration The total duration in seconds.
 	 * @param _feeSharing The address of fee sharing contract.
 	 * @param _vestingOwner The address of an owner of vesting contract.
+	 * @dev _vestingOwner should ALWAYS be multisig.
```
tjcloa, line 29: renaming to `_vestingOwnerMultisig` will make it more clear that multisig should be used

contracts/governance/Vesting/fouryear/FourYearVestingLogic.sol
```diff
@@ -137,6 +138,42 @@ contract FourYearVestingLogic is IFourYearVesting, FourYearVestingStorage, Appro
 		emit DividendsCollected(msg.sender, _loanPoolToken, _receiver, _maxCheckpoints);
 	}
 
+	/**
+	 * @notice Change token owner - only vesting owner.
```
tjcloa, line 142:
```suggestion
	 * @notice Change token owner - only vesting owner is allowed to change.
```

contracts/governance/Vesting/fouryear/FourYearVestingLogic.sol
```diff
@@ -137,6 +138,42 @@ contract FourYearVestingLogic is IFourYearVesting, FourYearVestingStorage, Appro
...
+	 * @dev Modifies token owner. This must be followed by approval
+	 * from token owner.
+	 * @param _newTokenOwner Address of new token owner.
```
tjcloa, line 145: should it be multisig only? if yes then pls rename the variable to _newTokenOwnerMultisig and at least add verification that the new address is a contract.

contracts/governance/Vesting/fouryear/FourYearVestingLogic.sol
```diff
@@ -137,6 +138,42 @@ contract FourYearVestingLogic is IFourYearVesting, FourYearVestingStorage, Appro
...
+	 * */
+	function changeTokenOwner(address _newTokenOwner) public onlyOwner {
```
tjcloa, line 147: auditors will yell: EXTERNAL! ) on all public

contracts/governance/Vesting/fouryear/FourYearVestingLogic.sol
```diff
@@ -137,6 +138,42 @@ contract FourYearVestingLogic is IFourYearVesting, FourYearVestingStorage, Appro
...
+	}
+
+	/**
+	 * @notice Approve token owner change - only token Owner.
+	 * @dev Token owner can only be modified
+	 * when both vesting owner and token owner have approved. This
+	 * function ascertains the approval of token owner.
+	 * */
+	function approveOwnershipTransfer() public onlyTokenOwner {
+		require(newTokenOwner != address(0), "invalid address");
```
tjcloa, line 160: not needed here as this is already verified when setting new token owner

contracts/governance/Vesting/fouryear/FourYearVestingLogic.sol
```diff
@@ -137,6 +138,42 @@ contract FourYearVestingLogic is IFourYearVesting, FourYearVestingStorage, Appro
...
+		emit TokenOwnerChanged(tokenOwner, msg.sender);
+	}
+
+	/**
+	 * @notice Set address of the implementation - only Token Owner.
+	 * @dev This function sets the new implementation address.
+	 * It must also be approved by the Vesting owner.
+	 * @param _newImplementation Address of the new implementation.
+	 * */
+	function setImpl(address _newImplementation) public onlyTokenOwner {
```
tjcloa, line 172: why implementation can be changed by the tokenOwner only?
form the task (story) statement:
```
Important:  both exchequer and investor need to confirm a contract update. exchequer could be the owner and the invesor the token owner.
```

contracts/governance/Vesting/fouryear/FourYearVestingFactory.sol
```diff
@@ -26,6 +26,7 @@ contract FourYearVestingFactory is IFourYearVestingFactory, Ownable {
...
 	 * @return The four year vesting contract address.
 	 * */
 	function deployFourYearVesting(
```
tjcloa, line 32: We will have to redeploy the factory should the logic change.
To avoid this we can pass the logic address as a param to the function or make the logic replaceable.

contracts/governance/Vesting/fouryear/FourYearVesting.sol
smbsp: Done

contracts/governance/Vesting/fouryear/FourYearVestingFactory.sol
smbsp: Done

contracts/governance/Vesting/fouryear/FourYearVestingLogic.sol
smbsp: Done

contracts/governance/Vesting/fouryear/FourYearVestingLogic.sol
smbsp: New token owner can be any other user

contracts/governance/Vesting/fouryear/FourYearVestingLogic.sol
smbsp: Changed to external

contracts/governance/Vesting/fouryear/FourYearVestingLogic.sol
smbsp: 1. setImpl() can only be called by the token owner. This function doesn't change implementation. It just modifies a state variable newImplementation.
2. The vesting owner(multisig) needs to call the setImplementation() with the same implementation contract address to modify the implementation/logic.

contracts/governance/Vesting/fouryear/FourYearVestingFactory.sol
smbsp: Done. This is an excellent point.

contracts/mockup/MockFourYearVestingLogic.sol
```diff
@@ -0,0 +1,12 @@
...
+import "../governance/Vesting/fouryear/FourYearVestingLogic.sol";
+
+contract MockFourYearVestingLogic is FourYearVestingLogic {
+	/**
+	 * @notice gets duration left
+	 */
+	function getDurationLeft() external view returns (uint256) {
+		return durationLeft;
+	}
+}
```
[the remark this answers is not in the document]
smbsp: The Mock contract is just to test upgradability in the tests. We won't require any more getters.

contracts/governance/Vesting/fouryear/FourYearVesting.sol
smbsp: Done

contracts/governance/Vesting/fouryear/FourYearVestingLogic.sol
```diff
@@ -126,6 +127,42 @@ contract FourYearVestingLogic is IFourYearVesting, FourYearVestingStorage, Appro
...
+		require(_newImplementation != address(0), "invalid new implementation address");
```
tjcloa, line 162:
```suggestion
		require(Address.isContract(_newImplementation), "invalid new implementation address");
```

tjcloa, line 162:
```suggestion
		require(Address.isContract(_newImplementation), "invalid new implementation address");
```

contracts/governance/Vesting/fouryear/FourYearVestingFactory.sol
```diff
@@ -10,31 +10,27 @@ import "./IFourYearVestingFactory.sol";
...
 		address _tokenOwner,
 		address _feeSharing,
-		address _vestingOwner
+		address _vestingOwnerMultisig,
+		address _fourYearVestingLogic
 	) external onlyOwner returns (address) {
-		address fourYearVesting = address(new FourYearVesting(fourYearVestingLogic, _SOV, _staking, _tokenOwner, _feeSharing));
-		Ownable(fourYearVesting).transferOwnership(_vestingOwner);
+		address fourYearVesting = address(new FourYearVesting(_fourYearVestingLogic, _SOV, _staking, _tokenOwner, _feeSharing));
+		Ownable(fourYearVesting).transferOwnership(_vestingOwnerMultisig);
```
tjcloa, line 33: what is the reason for this contract to exist? we can deploy the `FourYearVesting` contract directly. setting ownership transferring to the contract would be enough. also, for the testnet it is sometimes suboptimal to force transferring ownership to the multisig.

tjcloa, line 33: what is the reason for this contract to exist? we can deploy the `FourYearVesting` contract directly. setting ownership transferring to the contract would be enough. also, for the testnet it is sometimes suboptimal to force transferring ownership to the multisig.

contracts/governance/Vesting/fouryear/FourYearVestingLogic.sol
smbsp: I have kept a check here: https://github.com/DistributedCollective/Sovryn-smart-contracts/blob/1ae486d26377a2e25e2c81e4d4469f7cc2227dea/contracts/governance/Vesting/fouryear/FourYearVesting.sol#L63 We probably don't need this.

contracts/governance/Vesting/fouryear/FourYearVestingFactory.sol
smbsp: Also, testnet testing should not be a problem as the vesting contracts will be created for known token owner(my accounts)

contracts/mockup/MockFourYearVestingLogic.sol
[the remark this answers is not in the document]
tjcloa: this is not about mock contracts, sorry, need to have explained this.
this is a common issue when we would have to add getters when adding storage vars with `internal` acessibility after the initial deployment.
you cannot add public variables as they won't be visible as methods by proxy when delegating calls - only those registered explicitly as functions. 
so just to keep things consistent and uniform and clear it is better to have all the storage vars internal with getters when using proxy else half of vars will be public, the other one internal and there will be the third group which we do not need an access right away but when we do we can see poblic var which is not accessible.

contracts/governance/Vesting/fouryear/FourYearVestingLogic.sol
tjcloa: i think we can then remove verification completely from here as it will be done in the proxy anyway and it is not upgradeable so we are on the safe side

contracts/mockup/MockFourYearVestingLogic.sol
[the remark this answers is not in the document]
smbsp: I haven't managed time to do this due to several issues I faced while testnet testing. Hence, added a comment to make the new state variables internal and to add a getter function respectively. Sorry about going ahead with the dirty solution.

Although I think we can also make the new state variables `public` and add a getter with the same name. That would remove any inconsistency that you mentioned and also get the job done.

contracts/governance/Vesting/fouryear/FourYearVestingLogic.sol
smbsp: But we check for the address that is set here and hence put the zero address check. I haven't put the zero address check on the proxy.

contracts/governance/Vesting/fouryear/FourYearVestingFactory.sol
jameshowlett977: As you can see factory includes logic contract that can be relatively big, so using factory we separate bytecode of logic contact from bytecode of other contracts (e.g. VestingRegistry), otherwise we need to include bytecode of logic contract to VestingRegistry and can easily exceed 24 kb.

PR #423, opened 2022-03-21
Staking Security Update - implementation for SIP-0042

tjcloa:
Implementation of SIP-0042 https://github.com/DistributedCollective/SIPS/pull/42


contracts/governance/Staking/Staking.sol
```diff
@@ -269,7 +269,7 @@ contract Staking is IStaking, WeightedStaking, ApprovalReceiver {
 		uint96 amount,
 		uint256 until,
 		address receiver
-	) public {
+	) public whenNotFrozen {
```
smbsp, line 272: Do we need the `whenNotFrozen` check here? The `governanceWithdraw` function can only be executed by the exchequer.

The flow looks like:
governanceWithdrawVesting(Staking) -> governanceWithdrawTokens(Vesting) -> governanceWithdraw(Staking)

tjcloa: freeze function should be rarely applied - only when withdrawal is affected - exactly this case. in theory this can be vulnerable too.

contracts/governance/Staking/WeightedStaking.sol
```diff
@@ -25,6 +25,41 @@ contract WeightedStaking is Checkpoints {
...
+	/**
+	 * @dev Throws if paused.
+	 */
+	modifier whenNotPaused() {
+		require(!paused, "paused");
+		_;
+	}
+
+	/**
+	 * @dev Throws if paused.
```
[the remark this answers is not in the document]
tjcloa: fixed

contracts/governance/Staking/Staking.sol
HakuryuuHakuu: seconded. it is a failsafe measure, it has to apply everywhere to efficiently protect the contract.

eMarchenko (commented):
looks good.

you can probably save some space by delaying conversion to shorter uints to the latest possible moment, i.e. 

```js
pragma solidity 0.5.17;

library Math {
    function add96(uint96 x, uint96 y, string memory reason) internal pure returns (uint96) {
        require(x < x + y, reason);
        return x + y;
    }
}

library Cast {
    function cast96(uint x, string memory reason) internal pure returns (uint96) {
        require(x < 1<<96, reason);
        return uint96(x);
    }
}

contract A {
    uint96 public x;
    
    function foo(uint96 y) external {
        x = Math.add96(x, y, "error");
    }
}

contract B {    
    uint96 public x;
    
    function foo(uint y) external {
        x = Cast.cast96(x + y, "error");
    }
}
```
contract B is significantly cheaper than A. The difference increases for each short function parameter and checked arithmetic operation.

HakuryuuHakuu (commented):
I'm on the fence on this one, but someone needs to say it and I'll gladly play devil's advocate here:
I would argue that in the spirit of "fairness" and to be on the safe side (plus being true to what is written in sip42), we should fully lock the contract in case of freeze, meaning all public non-view functions that do not bear the whenNotPaused modifier already should bear whenNotFrozen. This correctly implements the idea of a full freeze on the whole contract, not putting anyone at an advantage, not leaving any door open to any kind of exploit.

HakuryuuHakuu (commented):
the current implementation allows all functions protected by whenNotPaused to be called when the state is frozen.
To fix this, either apply whenNotFrozen to functions that bear whenNotPaused, or change the whenNotPaused modifier to include a check on frozen (i.e. whenNotPaused becomes actually a shortname for whenNotPausedNorFrozen).
It is important for the notion of "freeze = pause + more locks" to be hardcoded. It is not safe enough to just add pause before freeze in a script file, we need the notion enforced by the smart contract.

contracts/governance/Staking/WeightedStaking.sol
```diff
@@ -589,6 +613,42 @@ contract WeightedStaking is Checkpoints {
...
+	 * @param _pause true when pausing, false when unpausing
+	 * */
+	function pauseUnpause(bool _pause) public onlyAuthorizedOrPauser {
+		paused = _pause;
+		emit StakingPaused(_pause);
+	}
+
+	/**
+	 * @notice Freeze contract - disable functions available even when paused
+	 * @param _freeze true when pausing, false when unpausing
```
HakuryuuHakuu, line 645: comment mismatch, switch pause for freeze where needed

Alitasovryn (approved):

contracts/governance/Staking/WeightedStaking.sol
```diff
@@ -589,6 +613,42 @@ contract WeightedStaking is Checkpoints {
...
+	 * */
+	function freezeUnfreeze(bool _freeze) public onlyAuthorizedOrPauser {
```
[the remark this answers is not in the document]
tjcloa: implemented this logic

contracts/governance/Staking/Staking.sol
tjcloa: changed logic and implementation

contracts/governance/Staking/WeightedStaking.sol
tjcloa: fixed

PR #429, opened 2022-04-07
Sip 0044

HakuryuuHakuu:
Please review mitigation strategy.
The idea is to thwart the whole family of exploits by disallowing (reverting on) certain operations in the same block as staking for the same unlock timestamp.


jameshowlett977 (approved):
The fix is ok, however I haven't found tests for cases:
`require(notSameBlock, "S20"); //S20 : "cannot be mined in the same block as last stake"`

HakuryuuHakuu:
> The fix is ok, however I haven't found tests for cases: `require(notSameBlock, "S20"); //S20 : "cannot be mined in the same block as last stake"`

Right, it misses a unit test for that.
The tests that show the exploit being thwarted are hitting that require though (or they would let the exploit run and the test would fail)
Nevertheless it is good practice to test ondividually. Will add this asap

HakuryuuHakuu:
> 

There are now 3 tests that show that very `require` breaking the transaction in the cases of:
-delegate
-extend
-withdraw
It also gets hit on each of the exploit regression tests. Since we are working inside a single block I did not find a proper way to get the solidity message out to check for "S20" explicitly, but those 3 elementary tests seem explicit enough.
The code is properly covered.

Alitasovryn (approved):

PR #433, opened 2022-04-29
Enable hardhat forking RSK mainnet

koirikivi:
Enable `npx hardhat node --fork` (and related test patterns) with RSK. Requires a bit of patching because of RSK.

The hardhat patch is automatically applied after running `npm install`

There's a command (`npm run fork:rsk-mainnet`) to start a local node forking from mainnet (`npm run fork:rsk-testnet` for the testnet).
Forking can be also used in tests. Please refer to Hardhat forking documentation https://hardhat.org/hardhat-network/guides/mainnet-forking#resetting-the-fork

## IMPORTANT!! Consider before merging
- This PR includes ~test~ hardhat task `check-fork-patch` to verify if the forking patch works. It will fail if there's no network connection or if the mainnet sovryn rpc node is down, which causes tests to fail.
- The patch will probably not work after hardhat is updated (we should get a patch upstream to hardhat, but it requires more work)
- ~This PR doesn't have the `package-lock.json` or `yarn.lock` modifications, because I'm not sure if we're using `npm` or `yarn` and which version of `node` and `npm` we're using~ `package-lock.json` is added too.
- ~There are some issues with `eth_getStorageAt` returning `0x0` -- a fix in progress.~ Should be fixed.
- The patch has **not** been tested extensively, though everything seems to work!

## Technical details

Calling a forked node (`hardhat node --fork MY_RSK_NODE_RPC_URL`) failed with:

    eth_call
    Invalid JSON-RPC response's result.  Errors: Invalid value null supplied
    to: RpcBlockWithTransactions | null/transactions: RpcTransaction
    Array/0: RpcTransaction/v: QUANTITY, Invalid value null supplied to:
    RpcBlockWithTransactions | null/transactions: RpcTransaction Array/0:
    RpcTransaction/r: QUANTITY, Invalid value null supplied to:
    RpcBlockWithTransactions | null/transactions: RpcTransaction Array/0:
    RpcTransaction/s: QUANTITY

We use `patch-package` to patch the local `hardhat` installation, since there's no upstream patch

This patch is based on:
https://gist.github.com/0x0scion/0422f9135bc37642ba36d55b59e8b424

More reading:
https://github.com/NomicFoundation/hardhat/issues/2395
https://github.com/NomicFoundation/hardhat/pull/2313/files
https://github.com/NomicFoundation/hardhat/issues/2106


tjcloa (commented):
lgtm
to work around possible issues with future hh updates, i suggest:
1. create a hh task from the forking.test.js and remove the test as it does not test the domain logic; this way we can avoid false-negatives due to the node issues etc.
2. add the PR comment as disclaimer and containing the instruction on removing the patch as .md doc; 
2.1. do a hook - overload hh task with initial warning with the doc link which can be removed once the patch is included in hh

koirikivi:
@tjcloa changes lgtm

dzarezenko:
Could you provide some instructions on how to apply this? I was trying to fork Harmony mainnet and received this error: 

```bash
 An unexpected error occurred:

InvalidResponseError: Invalid JSON-RPC response's result.

Errors: Invalid value undefined supplied to : RpcBlockWithTransactions | null/totalDifficulty: QUANTITY
```

PR #434, opened 2022-04-29
Refactoring/margin trading

cwsnt:


contracts/connectors/loantoken/LoanTokenLogicStandard.sol
```diff
@@ -1191,16 +1176,16 @@ contract LoanTokenLogicStandard is LoanTokenLogicStorage {
...
 		address _wrbtcToken = wrbtcTokenAddress;
 		address _loanTokenAddress = loanTokenAddress;
-		address receiver = sentAddresses[2];
-		uint256 newPrincipal = sentAmounts[1];
-		uint256 loanTokenSent = sentAmounts[3];
-		uint256 collateralTokenSent = sentAmounts[4];
+		address receiver = sentAddresses.receiver;
+		uint256 newPrincipal = sentAmounts.newPrincipal;
+		uint256 loanTokenSent = sentAmounts.loanTokenSent;
+		uint256 collateralTokenSent = sentAmounts.collateralTokenSent;
```
tjcloa, line 1188: no need to initialize extra variables - the struct vars can be used directly

contracts/modules/LoanOpenings.sol
```diff
@@ -375,6 +375,9 @@ contract LoanOpenings is LoanOpeningsEvents, VaultController, InterestUser, Swap
 			sentValues.collateralTokenSent = sentValues
 				.collateralTokenSent /// collateralTokenReceived
 				.add(receivedAmount);
+
+			/// Check the minEntryPrice with the rate
+			require(sentValues.loanTokenSent >= sentValues.minEntryPrice, "entry price above the minimum");
```
ororopickpocket: this is super confusing. I know that that variable was repurposed already before, but would be nicer to have another local variable holding it. On the first glance is looks awefully wrong, althought it's not. Just because of the naming.

ororopickpocket:
i've got following failing tests on that PR:

```
1) Contract: FeeSharingProxy:
       withdraw
         Should be able to withdraw (WRBTC pool):

      AssertionError: expected '9999842262857800000000' to equal '9999845829833800000000'
      + expected - actual

      -9999842262857800000000
      +9999845829833800000000
      
      at Context.<anonymous> (tests/FeeSharingProxyTest.js:1144:59)
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (internal/process/task_queues.js:93:5)

  2) Contract: FeeSharingProxy:
       withdraw
         Should be able to withdraw using 3 checkpoints:

      AssertionError: expected '9999842262896600000000' to equal '9999845829832600000000'
      + expected - actual

      -9999842262896600000000
      +9999845829832600000000
      
      at Context.<anonymous> (tests/FeeSharingProxyTest.js:1283:59)
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (internal/process/task_queues.js:93:5)

  3) Contract: GovernorAlpha (Any User Functions)
       "before all" hook: Initiating Accounts & Contracts for "Cannot execute proposal out of queue, can queue a successful proposal, emit ProposalQueued, execute it, emit ProposalExecuted, cannot remove it, and cannot execute it again":
     Error: VM Exception while processing transaction: revert Timelock::queueTransaction: Estimated execution block must satisfy delay.
      at Timelock.queueTransaction (contracts/governance/Timelock.sol:196)
      at TruffleContract.queueTransaction (node_modules/@nomiclabs/truffle-contract/lib/execute.js:169:26)
      at Context.<anonymous> (tests/Governance/GovernorAlpha/anyone.test.js:154:24)
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (internal/process/task_queues.js:93:5)
      at runNextTicks (internal/process/task_queues.js:62:3)
      at listOnTimeout (internal/timers.js:523:9)
      at processTimers (internal/timers.js:497:7)

  4) PoC
       Exploit:
     TransactionExecutionError: VM Exception while processing transaction: revert S04
      at HardhatNode._manageErrors (node_modules/hardhat/src/internal/hardhat-network/provider/node.ts:1241:14)
      at HardhatNode._gatherTraces (node_modules/hardhat/src/internal/hardhat-network/provider/node.ts:993:30)
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (internal/process/task_queues.js:93:5)
      at runNextTicks (internal/process/task_queues.js:62:3)
      at listOnTimeout (internal/timers.js:523:9)
      at processTimers (internal/timers.js:497:7)
      at HardhatNode._mineBlockWithPendingTxs (node_modules/hardhat/src/internal/hardhat-network/provider/node.ts:1070:23)
      at HardhatNode.mineBlock (node_modules/hardhat/src/internal/hardhat-network/provider/node.ts:369:16)
      at EthModule._sendTransactionAndReturnHash (node_modules/hardhat/src/internal/hardhat-network/provider/modules/eth.ts:1373:18)
      at HardhatNetworkProvider.request (node_modules/hardhat/src/internal/hardhat-network/provider/provider.ts:99:18)
      at EthersProviderWrapper.send (node_modules/@nomiclabs/hardhat-ethers/src/internal/ethers-provider-wrapper.ts:13:20)

  5) Contract: Staking
       Voting Power at first unstake time
         after shaking, does not give more voting power than tokens staked:

      AssertionError: expected '700' to equal '500'
      + expected - actual

      -700
      +500
      
      at Context.<anonymous> (tests/staking/VoteExploit.test.js:126:74)
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (internal/process/task_queues.js:93:5)

  6) Contract: Staking
       Voting Power shaking exploit
         sustains shaking in a single block:
     Transaction: 0xda89f62fdbeb50c93a1a7d2dedf46c50ba9c6b0949010fc144278c99bd776dcd exited with an error (status 0). 
     Please check that the transaction:
     - satisfies all conditions set by Solidity `require` statements.
     - does not trigger a Solidity `revert` statement.

  StatusError: Transaction: 0xda89f62fdbeb50c93a1a7d2dedf46c50ba9c6b0949010fc144278c99bd776dcd exited with an error (status 0). 
      at Context.<anonymous> (tests/staking/VoteExploitBoostedSimpler.test.js:196:38)
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (internal/process/task_queues.js:93:5)

```

same for you?

cwsnt:
> i've got following failing tests on that PR:
> 
> ```
> 1) Contract: FeeSharingProxy:
>        withdraw
>          Should be able to withdraw (WRBTC pool):
> 
>       AssertionError: expected '9999842262857800000000' to equal '9999845829833800000000'
>       + expected - actual
> 
>       -9999842262857800000000
>       +9999845829833800000000
>       
>       at Context.<anonymous> (tests/FeeSharingProxyTest.js:1144:59)
>       at runMicrotasks (<anonymous>)
>       at processTicksAndRejections (internal/process/task_queues.js:93:5)
> 
>   2) Contract: FeeSharingProxy:
>        withdraw
>          Should be able to withdraw using 3 checkpoints:
> 
>       AssertionError: expected '9999842262896600000000' to equal '9999845829832600000000'
>       + expected - actual
> 
>       -9999842262896600000000
>       +9999845829832600000000
>       
>       at Context.<anonymous> (tests/FeeSharingProxyTest.js:1283:59)
>       at runMicrotasks (<anonymous>)
>       at processTicksAndRejections (internal/process/task_queues.js:93:5)
> 
>   3) Contract: GovernorAlpha (Any User Functions)
>        "before all" hook: Initiating Accounts & Contracts for "Cannot execute proposal out of queue, can queue a successful proposal, emit ProposalQueued, execute it, emit ProposalExecuted, cannot remove it, and cannot execute it again":
>      Error: VM Exception while processing transaction: revert Timelock::queueTransaction: Estimated execution block must satisfy delay.
>       at Timelock.queueTransaction (contracts/governance/Timelock.sol:196)
>       at TruffleContract.queueTransaction (node_modules/@nomiclabs/truffle-contract/lib/execute.js:169:26)
>       at Context.<anonymous> (tests/Governance/GovernorAlpha/anyone.test.js:154:24)
>       at runMicrotasks (<anonymous>)
>       at processTicksAndRejections (internal/process/task_queues.js:93:5)
>       at runNextTicks (internal/process/task_queues.js:62:3)
>       at listOnTimeout (internal/timers.js:523:9)
>       at processTimers (internal/timers.js:497:7)
> 
>   4) PoC
>        Exploit:
>      TransactionExecutionError: VM Exception while processing transaction: revert S04
>       at HardhatNode._manageErrors (node_modules/hardhat/src/internal/hardhat-network/provider/node.ts:1241:14)
>       at HardhatNode._gatherTraces (node_modules/hardhat/src/internal/hardhat-network/provider/node.ts:993:30)
>       at runMicrotasks (<anonymous>)
>       at processTicksAndRejections (internal/process/task_queues.js:93:5)
>       at runNextTicks (internal/process/task_queues.js:62:3)
>       at listOnTimeout (internal/timers.js:523:9)
>       at processTimers (internal/timers.js:497:7)
>       at HardhatNode._mineBlockWithPendingTxs (node_modules/hardhat/src/internal/hardhat-network/provider/node.ts:1070:23)
>       at HardhatNode.mineBlock (node_modules/hardhat/src/internal/hardhat-network/provider/node.ts:369:16)
>       at EthModule._sendTransactionAndReturnHash (node_modules/hardhat/src/internal/hardhat-network/provider/modules/eth.ts:1373:18)
>       at HardhatNetworkProvider.request (node_modules/hardhat/src/internal/hardhat-network/provider/provider.ts:99:18)
>       at EthersProviderWrapper.send (node_modules/@nomiclabs/hardhat-ethers/src/internal/ethers-provider-wrapper.ts:13:20)
> 
>   5) Contract: Staking
>        Voting Power at first unstake time
>          after shaking, does not give more voting power than tokens staked:
> 
>       AssertionError: expected '700' to equal '500'
>       + expected - actual
> 
>       -700
>       +500
>       
>       at Context.<anonymous> (tests/staking/VoteExploit.test.js:126:74)
>       at runMicrotasks (<anonymous>)
>       at processTicksAndRejections (internal/process/task_queues.js:93:5)
> 
>   6) Contract: Staking
>        Voting Power shaking exploit
>          sustains shaking in a single block:
>      Transaction: 0xda89f62fdbeb50c93a1a7d2dedf46c50ba9c6b0949010fc144278c99bd776dcd exited with an error (status 0). 
>      Please check that the transaction:
>      - satisfies all conditions set by Solidity `require` statements.
>      - does not trigger a Solidity `revert` statement.
> 
>   StatusError: Transaction: 0xda89f62fdbeb50c93a1a7d2dedf46c50ba9c6b0949010fc144278c99bd776dcd exited with an error (status 0). 
>       at Context.<anonymous> (tests/staking/VoteExploitBoostedSimpler.test.js:196:38)
>       at runMicrotasks (<anonymous>)
>       at processTicksAndRejections (internal/process/task_queues.js:93:5)
> ```
> 
> same for you?

This is due too the hardhat version.
Please make sure you upgrade the hh version by  run the npm ci.

contracts/connectors/loantoken/LoanTokenLogicStandard.sol
cwsnt: Agree for the receiver.
But, for newPrincipal, lanTokenSent, collateralTokenSent is being used multiple in the rest of function execution, using struct var directly will require extra gas.

I will remove the receiver declaration, but will keep the rest as it is

contracts/modules/LoanOpenings.sol
cwsnt: The `sentValues.loanTokenSent` will be used by `_finalizeOpen` & `_emitOpeningEvents` after repurposed.
If we remove the "repurposed" approach and add new local variable holding it, meaning that we have to add new parameters in the `_finalizeOpen` & `_emitOpeningEvents`, which will require more gas.

contracts/connectors/loantoken/LoanTokenLogicStandard.sol
tjcloa: it is more clear which data is used when referring to the struct vars.  
what is advantage of using yet another local vars instead of using the struct var directly? why do you want to keep them?

contracts/modules/LoanOpenings.sol
```diff
@@ -85,8 +86,8 @@ contract LoanOpenings is
         bytes32 loanId,
         bool isTorqueLoan,
         uint256 initialMargin,
-        address[4] calldata sentAddresses,
-        uint256[5] calldata sentValues,
+        MarginTradeStructHelpers.SentAddresses calldata sentAddresses,
+        MarginTradeStructHelpers.SentAmounts calldata sentValues,
```
tjcloa, line 90: nat spec of the function does not correspond to the actual struct vars
```
    struct SentAmounts {
        uint256 interestRate;
        uint256 newPrincipal;
        uint256 interestInitialAmount;
        uint256 loanTokenSent;
        uint256 collateralTokenSent;
        uint256 minEntryPrice;
    }
    ```

File: contracts/connectors/loantoken/LoanTokenLogicStandard.sol (hunk 1 above)
Comment (cwsnt) [reply contracts/connectors/loantoken/LoanTokenLogicStandard.sol]: As i mentioned above, using local vars here is saving some gas

File: contracts/events/ProtocolSettingsEvents.sol [hunk 4]
```diff
@@ -159,4 +160,17 @@ contract ProtocolSettingsEvents is ModulesCommonEvents {
...
+        address indexed sender,
+        address indexed sourceTokenAddress,
+        address indexed destTokenAddress,
+        IERC20[] defaultPath
+    );
+
+    event RemoveDefaultPathConversion(
+        address indexed sender,
+        address indexed sourceTokenAddress,
+        address indexed destTokenAddress
```
tjcloa, line 174:
```suggestion
        address indexed destTokenAddress,
        IERC20[] defaultPath
```
to store the path removed

contracts/modules/LoanOpenings.sol
```diff
@@ -360,50 +361,58 @@ contract LoanOpenings is
...
+                    sentAddresses.borrower, /// borrower
                     loanLocal.id,
                     _collateralToken, /// fee token
                     _loanToken, /// pairToken (used to check if there is any special rebates or not) -- to pay fee reward
                     borrowingFee
                 );
 
-                sentValues[4] = sentValues[4] /// collateralTokenReceived
+                sentValues.collateralTokenSent = sentValues
+                    .collateralTokenSent /// collateralTokenReceived
```
tjcloa, line 389: pls consider adding a new var to the sentValues `collateralTokenReceived` or `sentAmountSubtraction`
which would add clarity at a price of another var in the struct.
separation of the amount sent and received (sent sub fees) this would lead to some code changes as not amountSent but amountReceived (or amountSent.sub(sentAmountSubtraction)) will be used further on

contracts/modules/LoanOpenings.sol
```diff
@@ -412,25 +421,25 @@ contract LoanOpenings is
...
                 collateralAmountRequired
             ),
             "collateral insufficient"
         );
 
-        loanLocal.collateral = loanLocal.collateral.add(sentValues[4]);
+        loanLocal.collateral = loanLocal.collateral.add(sentValues.collateralTokenSent);
 
         if (isTorqueLoan) {
             /// reclaiming varaible -> interestDuration
```
tjcloa, line 433:
```suggestion
            /// reclaiming variable -> interestDuration
```
the code below is a great confusion
pls consider adding another variable for TorqueLoan and amend the comment above respectively

contracts/modules/ProtocolSettings.sol
```diff
@@ -863,4 +866,75 @@ contract ProtocolSettings is
...
+        address destTokenAddress,
+        IERC20[] calldata defaultPath
+    ) external onlyOwner whenNotPaused {
+        require(
+            Address.isContract(sourceTokenAddress) && Address.isContract(destTokenAddress),
+            "ERR_NON_CONTRACT_ADDR"
+        );
+
+        uint256 defaultPathLength = defaultPath.length;
+        require(defaultPathLength >= 3, "ERR_PATH_LENGTH");
```
tjcloa, line 905:
```suggestion
        require(defaultPathLength > 0, "ERR_PATH_LENGTH");
```
the first and the last elements in the defaultPath are always source and dest tokens
consider passing only the path between them to remove duplication when passing them also in the defaultPath param.

contracts/swaps/connectors/SwapsImplSovrynSwap.sol
```diff
@@ -84,10 +84,20 @@ contract SwapsImplSovrynSwap is State, ISwapsImpl {
...
         ISovrynSwapNetwork sovrynSwapNetwork =
             getSovrynSwapNetworkContract(sovrynSwapContractRegistryAddress);
+
+        IERC20[] memory _defaultPathConversion =
+            defaultPathConversion[sourceTokenAddress][destTokenAddress];
+
+        /// will use the defaultPath if it's set, otherwise query from the SovrynSwapNetwork.
         IERC20[] memory path =
-            sovrynSwapNetwork.conversionPath(IERC20(sourceTokenAddress), IERC20(destTokenAddress));
+            _defaultPathConversion.length >= 3
```
tjcloa, line 93:
```suggestion
            _defaultPathConversion.length > 0
```
the same suggestion as in the ProtocolSettings contract

contracts/swaps/connectors/SwapsImplSovrynSwap.sol
```diff
@@ -107,11 +117,6 @@ contract SwapsImplSovrynSwap is State, ISwapsImpl {
                 "insufficient source tokens provided."
             );
             minReturn = requiredDestTokenAmount;
-        } else if (sourceTokenAmountUsed > 0) {
-            /// For some reason the Sovryn swap network tends to return a bit less than the expected rate.
-            minReturn = sovrynSwapNetwork.rateByPath(path, sourceTokenAmountUsed).mul(995).div(
-                1000
-            );
```
tjcloa, line 114: is this not an issue anymore?

tests/loan-token/TradingTestToken.test.js
```diff
@@ -792,5 +792,287 @@ contract("LoanTokenTrading", (accounts) => {
...
+                SUSD.address,
+                new BN(10).pow(new BN(20)).toString()
+            );
+            await priceFeeds.setRates(
+                RBTC.address,
+                SUSD.address,
+                new BN(10).pow(new BN(20)).toString()
+            );
+
+            const defaultPathConversion = [SUSD.address, RBTC.address, RBTC.address];
```
tjcloa, line 828: the conversion path should not have the same token as the source and dest

contracts/modules/ProtocolSettings.sol
```diff
@@ -863,4 +866,75 @@ contract ProtocolSettings is
...
+    function setDefaultPathConversion(
+        address sourceTokenAddress,
+        address destTokenAddress,
+        IERC20[] calldata defaultPath
+    ) external onlyOwner whenNotPaused {
+        require(
+            Address.isContract(sourceTokenAddress) && Address.isContract(destTokenAddress),
+            "ERR_NON_CONTRACT_ADDR"
+        );
+
```
tjcloa, line 903: pls add require the defaultPath contains neither source nor dest tokens

contracts/modules/LoanOpenings.sol
cwsnt: I don't think adding variables here is efficient since the data is being used actually the same, it's just a matter of term which is different between the protocol modules (collateralTokenReceived) and the loan token logic (collateralTokenSent), but the value is the same for both.
So we can just unify the terms for both protocol and loan token logic

contracts/modules/ProtocolSettings.sol
cwsnt: Yes, the source & dest tokens would in the first & last index, but in the case of the minimum path length, it will always be 3 at minimum.

contracts/swaps/connectors/SwapsImplSovrynSwap.sol
cwsnt: Yes, querying minReturn to the AMM at the same block is kind of useless, so we chnage it to 1 instead as the minReturn

tests/loan-token/TradingTestToken.test.js
cwsnt: Yeah, this is just a use for test purposes, in the mainnet, shouldn't be the case as in it will have 3 length of addresses at very minimum (especially for v1 pool)

contracts/modules/ProtocolSettings.sol
tjcloa: > pls consider passing only the path between them to remove duplication when passing them also in the defaultPath param.

contracts/swaps/connectors/SwapsImplSovrynSwap.sol
tjcloa: i mean https://github.com/DistributedCollective/Sovryn-smart-contracts/pull/434#discussion_r897369137

tests/loan-token/TradingTestToken.test.js
tjcloa: i mean i believe it should be restricted
else in case of an error it will work extremely inefficient and pretty hard to discover

contracts/modules/LoanOpenings.sol
tjcloa: @ororopickpocket wdyt?

contracts/modules/LoanOpenings.sol
cwsnt: Will add new property in the struct

contracts/modules/LoanOpenings.sol
cwsnt: Will remove the comment

contracts/modules/ProtocolSettings.sol
cwsnt: Need to remove the source & dest token address params, use the first & n length array instead

cwsnt:
We need to consider if the internal swap is called directly (without delegatecall) --> In this case, it can't read the defaultPath storage in the protocol.
1. Solutions:
- Remove all of the direct call (change it into delegateCall)
- Pass defaultPath as the parameter in the SwapsSovrynImpl

2. Need to use the defaultPath in internalExpectedRate & internalExpectedReturn

cwsnt:
> * the

`_swapsExpectedReturn` is calling the `internalExpectedReturn` function in SovrynSwapImpl directly, and then most of function that call `_swapsExpectedReturn`  has view visibility, so changing into delegateCall will not effective in here as in we need to remove the view visibility which will cause another issues.

So, lets go with the second solution.

contracts/swaps/connectors/SwapsImplSovrynSwap.sol
```diff
@@ -253,13 +246,38 @@ contract SwapsImplSovrynSwap is State, ISwapsImpl {
...
+
         IERC20[] memory path =
-            sovrynSwapNetwork.conversionPath(IERC20(sourceTokenAddress), IERC20(destTokenAddress));
+            defaultPath.length >= 3
+                ? defaultPath
+                : sovrynSwapNetwork.conversionPath(
+                    IERC20(sourceTokenAddress),
+                    IERC20(destTokenAddress)
+                );
+
```
ororopickpocket, line 262: why aren't you using getConversionPath here as well?

ororopickpocket:
@Alitasovryn QA for this PR can start.
This is a refactoring PR, so there are no new features which would require new test cases. The most important thing is that the old tests are still passing. The only new tests required are for adding the default path.
After testnet deployment, margin trading should be tested with the dapp.

contracts/swaps/connectors/SwapsImplSovrynSwap.sol
ororopickpocket: If we go forward with this, there should be at least a warning on the documentation. It might be cleaner to implement this in a way that none of the functions assumes to have access to the protocol storage.
@tjcloa what do you think?

Alitasovryn:
@***@***.*** ***@***.***>  Thank you Ororo.
Will look into it.

contracts/swaps/connectors/SwapsImplSovrynSwap.sol
cwsnt: Correct, this is the only function gets called directly instead of delegate call (by the `_swapsExpectedReturn` function).
I agree with adding the better & clearer comment for this part.
wdyt @tjcloa  ?

contracts/modules/LoanOpenings.sol
```diff
@@ -639,26 +647,27 @@ contract LoanOpenings is
...
      * @return The loanId.
      * */
     function _initializeLoan(
         LoanParams memory loanParamsLocal,
         bytes32 loanId,
         uint256 initialMargin,
-        address[4] memory sentAddresses,
-        uint256[5] memory sentValues
+        MarginTradeStructHelpers.SentAddresses memory sentAddresses,
+        MarginTradeStructHelpers.SentAmounts memory sentValues
```
tjcloa, line 663: we only need `newPrincipal` from  `MarginTradeStructHelpers.SentAmounts` - can be replaced with only one value

contracts/modules/LoanOpenings.sol
```diff
@@ -360,50 +363,53 @@ contract LoanOpenings is
...
+                    sentAddresses.borrower, /// borrower
                     loanLocal.id,
                     _collateralToken, /// fee token
                     _loanToken, /// pairToken (used to check if there is any special rebates or not) -- to pay fee reward
                     borrowingFee
                 );
 
-                sentValues[4] = sentValues[4] /// collateralTokenReceived
-                    .sub(borrowingFee);
+                sentValues.collateralTokenSent = sentValues.collateralTokenSent.sub(borrowingFee);
```
tjcloa, line 390: adding `collateralTokenReceived` to the struct will make the code much clearer
```
sentValues.collateralTokenReceived = sentValues.collateralTokenSent.sub(borrowingFee);
```
which also requires minor code changes in this function to use collateralTokenReceived as needed but will increase readability

contracts/modules/LoanOpenings.sol
```diff
@@ -511,11 +518,12 @@ contract LoanOpenings is
...
      *     manager: delegated manager of loan unless address(0).
      * @param sentValues The values to send:
-     *     newRate: New loan interest rate.
+     *     interestRate: New loan interest rate.
      *     newPrincipal: New loan size (borrowAmount + any borrowed interest).
-     *     torqueInterest: New amount of interest to escrow for Torque loan (determines initial loan length).
+     *     interestInitialAmount: New amount of interest to escrow for Torque loan (determines initial loan length).
      *     loanTokenReceived: Total loanToken deposit (amount not sent to borrower in the case of Torque loans).
-     *     collateralTokenReceived: Total collateralToken deposit.
+     *     collateralTokenSent: Total collateralToken deposit.
```
tjcloa, line 525: if `collateralTokenReceived` will be added to the struct then natspec comments shoud be raplaced respectively
```
*     collateralTokenReceived: Total collateralToken deposit.
```

contracts/modules/LoanOpenings.sol
```diff
@@ -549,17 +557,17 @@ contract LoanOpenings is
...
-                sentAddresses[0], /// lender
+                sentAddresses.borrower, /// user (trader)
+                sentAddresses.lender, /// lender
                 loanLocal.id, /// loanId
                 loanParamsLocal.collateralToken, /// collateralToken
                 loanParamsLocal.loanToken, /// loanToken
-                sentValues[4], /// positionSize
-                sentValues[1], /// borrowedAmount
-                sentValues[0], /// interestRate,
+                sentValues.collateralTokenSent, /// positionSize
```
tjcloa, line 565:
```
sentValues.collateralTokenSent, /// newCollateral
sentValues.collateralTokenReceived, /// newCollateral add converted loan to collateral amount
```

contracts/swaps/connectors/SwapsImplSovrynSwap.sol
tjcloa: as a quick solution i'm fine with adding clearer comments
but i agree with ororo that this design is in general confusing and can cause problems in the future
much clearer is to use logic contracts without caring abt their deployment addresses at all presuming it is used as a library with proxy storage only and all the calls are done via the proxy
i would prefer to either (on per-function basis):
- extract "direct" public functions like this into a separate contract or move it into an existing contract 
- use via delegatecall
depending on priorities and efforts to impelement this (which will require adapting tests too) a separate refactoring task can be added to monday.com
@ororopickpocket, @cwsnt wdyt?

contracts/swaps/connectors/SwapsImplSovrynSwap.sol
```diff
@@ -227,8 +223,10 @@ contract SwapsImplSovrynSwap is State, ISwapsImpl {
     ) public view returns (uint256) {
         ISovrynSwapNetwork sovrynSwapNetwork =
             getSovrynSwapNetworkContract(sovrynSwapContractRegistryAddress);
```
tjcloa, line 225: it shadows existing variable 

all params can be added leading underscores

contracts/swaps/SwapsUser.sol
```diff
@@ -259,7 +259,8 @@ contract SwapsUser is State, SwapsEvents, FeesHelper {
             sourceToken,
```
tjcloa, line 259: this call might need to be changed depending on what we decide on using swapsImpl directly (not via delegatecall from the protocol)

contracts/swaps/connectors/testnet/SwapsImplLocal.sol
```diff
@@ -99,14 +99,16 @@ contract SwapsImplLocal is State, ISwapsImpl {
...
+     * @param defaultPath defaultPath for swap.
      *
      * @return precision The expected return.
      * */
     function internalExpectedReturn(
         address sourceTokenAddress,
         address destTokenAddress,
         uint256 sourceTokenAmount,
-        address unused
+        address unused,
```
tjcloa, line 110: do we need to keep this param?

tjcloa (commented):
added inline comments

contracts/modules/LoanOpenings.sol
cwsnt: We have discussed this, it's just a matter of inaccurrate comment. So adding new variable to the struct for the same purpose is not really good idea

contracts/swaps/connectors/testnet/SwapsImplLocal.sol
cwsnt: Yes, to override the inheritted function from the parent

contracts/swaps/connectors/SwapsImplSovrynSwap.sol
cwsnt: Changed the variable name for better unification

contracts/modules/LoanOpenings.sol
cwsnt: Good idea

PR #435, opened 2022-05-05
Lm rewards refactor

smbsp:


contracts/farm/LockedSOVRewardTransferLogic.sol
```diff
@@ -0,0 +1,78 @@
...
+    function transferReward(
+        address _to,
+        uint256 _value,
+        bool _isWithdrawal
+    ) external {
+        IERC20 token = IERC20(lockedSOV.SOV());
+        require(token.transferFrom(msg.sender, address(this), _value), "Transfer failed");
+        require(token.approve(address(lockedSOV), _value), "Approve failed");
+        lockedSOV.deposit(_to, _value, unlockedImmediatelyPercent);
+        if (!_isWithdrawal) {
```
smbsp, line 74: Shouldn't it be `if (_isWithdrawal) {` - As per the definition in interface
`@param _isWithdrawal If true, means that the reward and the LP deposited tokens are being compeltely withdrawn`

contracts/farm/ERC20TransferLogic.sol
```diff
@@ -0,0 +1,41 @@
+pragma solidity 0.5.17;
+
+import "./IRewardTransferLogic.sol";
```
smbsp, line 3: Not required - as it has already been inherited in Storage

contracts/farm/LiquidityMiningStorageV1.sol
```diff
@@ -0,0 +1,18 @@
+pragma solidity 0.5.17;
```
smbsp, line 1: We can discard this file and add the state variables to LiquidityMiningStorage

contracts/farm/LiquidityMiningV1.sol
```diff
@@ -0,0 +1,849 @@
...
+    }
+
+    /* Functions */
+
+    /**
+     * @notice Initialize mining.
+     *
+     * @param _liquidityMiningV2 The LiquidityMiningV2 contract address
+     */
+    function initialize(address _liquidityMiningV2) external onlyAuthorized {
```
smbsp, line 83: I would prefer to rename this function to `setLiquidityMiningV2Address`

contracts/farm/LiquidityMiningProxyV2.sol
```diff
@@ -0,0 +1,11 @@
+pragma solidity ^0.5.17;
+
+import "./LiquidityMiningStorageV2.sol";
+import "../proxy/UpgradableProxy.sol";
+
+/**
+ * @dev LiquidityMining contract should be upgradable, use UpgradableProxy
+ */
+contract LiquidityMiningProxyV2 is LiquidityMiningStorageV2, UpgradableProxy {
```
smbsp, line 9: Can we update the LM address in RBTCWrapperProxy?

PR #444, opened 2022-06-24
fix WeightedStaking is Checkpoints cyclic reference

tjcloa:
No more `Definition of base has to precede definition of derived contract` on Checkpoints in 
`contract WeightedStaking is Checkpoints`.
A bit of optimization - using interfaces as types instead of contracts.


ororopickpocket (approved):

cwsnt (approved):
lgtm

PR #446, opened 2022-07-01
fix: less claimable SOV after stopblock

cwsnt:


tests/stakingRewards/StakingRewards.js
```diff
@@ -279,6 +280,42 @@ contract("StakingRewards", (accounts) => {
             expect(afterBalance).to.be.bignumber.greaterThan(beforeBalance);
         });
 
+        it("should getting correct rewards after stopblock", async () => {
```
tjcloa, line 283:
```suggestion
        it("should be getting correct rewards after stopblock", async () => {
```

tests/stakingRewards/StakingRewards.js
```diff
@@ -44,6 +44,7 @@ const {
 } = require("../Utils/initializer.js");
 
 const StakingLogic = artifacts.require("StakingMock");
+const StakingLogicMockup = artifacts.require("StakingMockup");
```
tjcloa, line 47: looks like it is better to move added weight by block mocking functionality to StakingMock

tjcloa: decided to postpone for refactoring

tests/stakingRewards/StakingRewards.js
tjcloa: fixed

PR #453, opened 2022-10-11
Feat: Add global reentrancy guard & invariant check

cwsnt:
### Background information
There are two relevant smart contracts involved in the exploit. One is the lending pool aka loan token contract, which has the public endpoints for borrowing, margin trading, minting (providing liquidity) and burning (removing liquidity). The other one is the lending, borrowing and margin trading protocol, which provides the functionality to open, maintain and close positions. For simplicity, we call the latter contract “the protocol” going forward. 

Those two contracts interact with each other. Borrowing and margin trading create positions on the protocol. Positions are being closed on the protocol, which returns the principal and interest to the lending pool.

The LP token price of the lending pools is defined by the ratio of the underlying asset to the total LP token supply. At the beginning the two are equal, but as interest is generated the number of underlying assets grows, changing the ratio and therefore increasing the LP token price. 

The underlying assets are held partially on the lending pool, partially on the protocol. The register of the funds currently borrowed out is held by the protocol, the non borrowed assets are held by the lending pool. When reading the total asset supply on the lending pool in order to calculate the price, it needs to read the protocol’s state as well. This is what makes the following exploit possible.

### Affected contracts
The pools affected by the exploit are the iRBTC and iUSDT lending pools. In the following section, we explain the exploit giving the example of the iRBTC pool. The exploit on iUSDT used the same underlying vulnerability but worked slightly differently.

### The exploit
Based on information gathered so far, the exploit was conducted in the following steps:
1. The attacker borrowed RBTC from the iRBTC lending pool with XUSD collateral.
2. The attacker closed the position by swapping part of its XUSD collateral for RBTC, but chose to receive back the remaining collateral (after repaying the loan) in XUSD. There is collateral remaining, because positions on the protocol need to be overcollateralized.
3. There is no way to buy a certain amount of RBTC on the Sovryn AMM. So, the protocol estimates how many XUSD it needs to sell to receive just enough RBTC to cover the loan. For small positions this difference is almost unnoticeable, but the bigger the position, the bigger the excess of RBTC received on the swap. If it exceeds a certain threshold, it is refunded to the borrower directly. 
This refund happens after the state was updated on the protocol, but before the principal was actually returned to the lending pool. 
4. The refund called the fallback function (default function triggered when receiving RBTC) on the attackers contract which uses the opportunity to mint LP tokens on the iRBTC lending pool. As mentioned above, the LP token price of the lending pools is defined by the ratio of the underlying asset to the total LP token supply. Because the protocol state was already updated (the number of assets belonging to the lending pool was reduced), but the tokens did not yet arrive at the lending pool, the calculated underlying asset balance of the lending pool was too low and the token price was lower than its actual value. Therefore, the attacker was able to purchase iTokens “at a discount”.
5. After minting the tokens, the execution of the closing function resumed, transferring the principal to the lending pool. This increased its underlying asset balance and consequently the token price.
6. Finally, the attacker burned his LP tokens with a profit.

We decided to call this a cross contract reentrancy attack. It is not a classical reentrancy in the sense that the attacker executes a function on the same contract. The protocol as well as the lending pools are guarded against reentrancy individually. This attack was only possible because the attacker interacted with two different contracts with separate storage, where one depended on the other. 

### The iUSDT case
The iRBTC scenario worked because there was a native currency transfer to the attacker, which triggered the fallback function. Regular ERC20 transfers do not notify the receiver and therefore do not offer the same opportunity. This is the reason why the DoC, BPro and XUSD lending pools remained untouched. 

iUSDT is different, because it is an ERC777 token, which calls the receiver when tokens are transfered. This allowed the attacker to perform the same attack with some small differences. 

The attacker borrowed iUSDT with RBTC collateral and closed the loan in the same transaction with a deposit. So, in contrast to the scenario above there was no swap involved, but the principal was returned by the borrower in the same currency. The transfer of RUSDT notified the sending contract before actually processing the transfer. This allowed the attacker to mint iTokens “at a discount” in the same fashion as described above.

### The remediation

We decided to implement two independent protections against this kind of attack. The first is a global reentrancy protection, which is not limited to securing one contract, but a system of interconnected contracts. The second is the introduction of invariant checks to the closing functions. We have been discussing the benefits of invariant checks for our smart contract security internally for a while and see the current situation as an opportunity to start introducing them.

#### The global reentrancy protection

The heart of the global reentrancy protection is a simple mutex contract which holds a counter and exposes a function to increment the counter and return its value. A shared reentrancy guard contract defines a modifier which increments the mutex counter in the beginning and verifies at the end of the function execution if the value returned by the mutex remains unchanged. If another function, which is protected in the same way, is called from within a protected function, it increments the counter and therefore leads to a revert at the end of the first function execution. With this solution, it is possible to prevent nested calls of protected functions, but still allow their subsequent execution.

#### The invariant check

The invariant check provides an additional layer of protection. It was added to both closing functions of the protocol. At the beginning of the function execution it reads the total supply of the lending pool’s LP tokens and at the end, it ensures that the total supply remains unchanged.


contracts/connectors/loantoken/LoanTokenLogicStandard.sol
```diff
@@ -96,6 +97,7 @@ contract LoanTokenLogicStandard is LoanTokenLogicStorage {
     function burn(address receiver, uint256 burnAmount)
         external
         nonReentrant
+        globallyNonReentrant
```
ororopickpocket, line 100: i think we should also add this to borrow and margin trade

contracts/modules/LoanClosingsShared.sol
```diff
@@ -319,28 +321,32 @@ contract LoanClosingsShared is
...
+        /// Cache the Loan Token Price
+        address loanTokenPoolAddress = underlyingToLoanPool[loanParamsLocal.loanToken];
+        require(loanTokenPoolAddress != address(0), "Invalid loan token pool address");
+
+        TempLoanPoolBalance memory loanPoolBalance =
+            TempLoanPoolBalance({
+                initialLoanTokenSupply: ILoanTokenModules(loanTokenPoolAddress).totalSupply(),
+                latestLoanTokenSupply: 0
+            });
+
```
ororopickpocket, line 349: could you move the invariant checks to a modifier, please?

````
modifier iTokenSupplyUnchanged(address loanId){
    check initial balance
    _;
    ensure that final balance == initial balance
}
````
this would make the whole thing much easier readable and reusable.

File: contracts/modules/LoanClosingsWith.sol [hunk 3]
```diff
@@ -104,14 +106,20 @@ contract LoanClosingsWith is LoanClosingsShared {
         )
     {
         _checkAuthorized(loanId);
-        return
+        CloseWithSwapReturn memory closeWithSwapReturn =
```
ororopickpocket, line 109: why did you introduce the CloseWithSwapReturn struct? Is it because of the invariant check?
I think it does not help readability. By moving the invariant checks out of the actual closing function you're solving the stack too deep problem in a nicer way.

contracts/modules/LoanClosingsWith.sol
```diff
@@ -145,6 +153,16 @@ contract LoanClosingsWith is LoanClosingsShared {
...
 
+        /// Cache the Loan Token Price
+        address loanTokenPoolAddress = underlyingToLoanPool[loanParamsLocal.loanToken];
+        require(loanTokenPoolAddress != address(0), "Invalid loan token pool address");
+
+        TempLoanPoolBalance memory loanPoolBalance =
+            TempLoanPoolBalance({
+                initialLoanTokenSupply: ILoanTokenModules(loanTokenPoolAddress).totalSupply(),
+                latestLoanTokenSupply: 0
+            });
```
ororopickpocket, line 164: same here. could reuse the same modifier. please avoid copying the same code to different functions. always use internal functions or modifiers for code used in multiple places.

ororopickpocket (changes requested):
This is part one of my review only. I'm already giving this feedback, so @cwsnt can work on the changes while i finish the rest.

contracts/connectors/loantoken/LoanTokenLogicStandard.sol
ororopickpocket: do i just not see it or are there indeed no global reentrancy checks on the closing functions of the protocol?

contracts/modules/LoanClosingsShared.sol
cwsnt: Moving it to modifier might cost another gas, since we need to read the loan & loanParams from the storage, which kinda redundant since we do it inside the function as well.
Wdyt?

contracts/modules/LoanClosingsWith.sol
cwsnt: There was stack too deep issue, not even 1 local variable can be added.

ororopickpocket: the additional gas cost shouldn't be too high and the improved readability and reusability is worth it, i think.

contracts/connectors/loantoken/LoanTokenLogicStandard.sol
cwsnt: Reentrancy checks has been added to closeWithSwap, closeWithDeposit, liquidate & rollover

contracts/testhelpers/TestDiscRBTC.sol
```diff
@@ -0,0 +1,130 @@
...
+ *
+ * The refund happened in step #3, which will send back the RBTC back to this contract.
+ * Then, this contract will try to do another iWRBTC minting to the loan token --> this is where the cross-reentrancy happened between the protocol & the loan token contract.
+ *
+ * This function should never been passed in the unit testing since we have:
+ * 1. invariant check for the loan token (iToken) total supply for closeWithSwap function.
+ * 2. global reentrancy guard between the protocol & the loan token.
+ */
+
+contract TestDiscRBTC {
```
tjcloa, line 27: pls rename the contract to `TestCrossReentrancyRBTC`

contracts/testhelpers/TestDiscRBTC.sol
```diff
@@ -0,0 +1,130 @@
...
+        bytes32 loanParamsLocalId =
+            ILoanTokenModules(loanTokenWRBTC).loanParamsIds(
+                uint256(keccak256(abi.encodePacked(SUSD, true)))
+            );
+        bytes32 loan_id =
+            keccak256(
+                abi.encodePacked(loanParamsLocalId, loanTokenWRBTC, _borrower, _borrowerNonce)
+            );
+
+        // STEP 3 close the borrowed position with a swap (probably works just as well with deposit)
```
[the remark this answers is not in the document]
tjcloa: i'll have a deeper look into it - no need to take any actions rn

tests/FeeSharingProxyTest.js
```diff
@@ -125,6 +126,8 @@ contract("FeeSharingProxy:", (accounts) => {
     });
 
     async function protocolDeploymentFixture(_wallets, _provider) {
+        await mutexUtils.getOrDeployMutex();
```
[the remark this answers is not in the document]
tjcloa: as discussed - it is needed for using in deployed contracts during the tests execution.  
pls add a short description why it is here.

tjcloa (changes requested):
required changes have been discussed with @cowsant

contracts/modules/LoanClosingsShared.sol
cwsnt: Updated to modifier

contracts/modules/LoanClosingsShared.sol
```diff
@@ -428,18 +444,31 @@ contract LoanClosingsShared is
...
 
+        /// Validate the Loan Token Price
+        loanPoolBalance.latestLoanTokenSupply = ILoanTokenModules(loanTokenPoolAddress)
+            .totalSupply();
+
+        require(
+            loanPoolBalance.initialLoanTokenSupply == loanPoolBalance.latestLoanTokenSupply,
+            "mismatch loan token supply"
+        );
+
```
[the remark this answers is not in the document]
cwsnt: Updated to modifier

contracts/modules/LoanClosingsWith.sol
cwsnt: Updated

contracts/modules/LoanClosingsWith.sol
cwsnt: Updated to modifier

tjcloa (changes requested):
required changes have been discussed with @cwsnt

contracts/testhelpers/TestDiscERC777.sol
```diff
@@ -0,0 +1,139 @@
...
+            _borrower,
+            _receiver,
+            ""
+        );
+
+        // step2 mint iSUSD
+        // prerequisite: this contract should have SUSD
+        uint256 _SUSDBalance = IERC20(SUSD).balanceOf(address(this));
+        IERC20(SUSD).approve(loanToken, _SUSDBalance);
+        ILoanTokenModules(loanToken).mint(_receiver, 100 ether);
```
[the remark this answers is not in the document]
cwsnt: Added the description in the test contract, and also refactor the testing to overwrite the modifier

contracts/interfaces/IERC1820Registry.sol
```diff
@@ -0,0 +1,126 @@
...
+ * Implementers may be shared by multiple accounts, and can also implement more
+ * than a single interface for each account. Contracts can implement interfaces
+ * for themselves, but externally-owned accounts (EOA) must delegate this to a
+ * contract.
+ *
+ * {IERC165} interfaces can also be queried via the registry.
+ *
+ * For an in-depth explanation and source code analysis, see the EIP text.
+ */
+interface IERC1820Registry {
```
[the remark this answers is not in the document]
cwsnt: Updated in https://github.com/DistributedCollective/Sovryn-smart-contracts/pull/453/commits/9174699f93acf8c4458d4c552b1180f82fb45214

contracts/testhelpers/TestDiscERC777.sol
```diff
@@ -0,0 +1,132 @@
...
+
+import "../interfaces/ILoanTokenModules.sol";
+import "../interfaces/IERC20.sol";
+import "../connectors/loantoken/interfaces/ProtocolLike.sol";
+import "../openzeppelin/SafeMath.sol";
+import "../interfaces/IWrbtcERC20.sol";
+import "../interfaces/IERC1820Registry.sol";
+import "../mockup/MockLoanTokenLogic.sol";
+
+contract TestDiscERC777 {
```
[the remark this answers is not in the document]
cwsnt: Updated in this [commit](https://github.com/DistributedCollective/Sovryn-smart-contracts/pull/453/commits/9174699f93acf8c4458d4c552b1180f82fb45214)

contracts/testhelpers/TestDiscRBTC.sol
```diff
@@ -0,0 +1,130 @@
...
+        sovrynProtocol.closeWithSwap(
+            loan_id,
+            msg.sender,
+            collateralTokenSent.mul(200).div(100), // make it 20% higher from initial collateral sent to make sure whole position is closed
+            true,
+            ""
+        );
+
+        /** Rest of code Should not be executed */
```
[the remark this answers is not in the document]
cwsnt: Updated in this [commit](https://github.com/DistributedCollective/Sovryn-smart-contracts/pull/453/commits/9174699f93acf8c4458d4c552b1180f82fb45214)

contracts/testhelpers/TestDiscRBTC.sol
```diff
@@ -0,0 +1,130 @@
...
+        // STEP 4 Burn all iRBTC
+        uint256 _iWRBTCBalance = ILoanTokenModules(loanTokenWRBTC).balanceOf(_borrower);
+        ILoanTokenModules(loanTokenWRBTC).burn(_receiver, _iWRBTCBalance);
+
+        balanceState memory finalBalance =
```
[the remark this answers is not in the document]
cwsnt: Commented out in this [commit](https://github.com/DistributedCollective/Sovryn-smart-contracts/pull/453/commits/9174699f93acf8c4458d4c552b1180f82fb45214)

tests/FeeSharingProxyTest.js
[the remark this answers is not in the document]
cwsnt: Updated in this [commit](https://github.com/DistributedCollective/Sovryn-smart-contracts/pull/453/commits/9174699f93acf8c4458d4c552b1180f82fb45214)

tests/loan-token/BorrowingTestToken.test.js
```diff
@@ -17,10 +17,18 @@ const { loadFixture } = waffle;
 const { expectRevert, BN, expectEvent } = require("@openzeppelin/test-helpers");
 const FeesEvents = artifacts.require("FeesEvents");
 const LoanOpenings = artifacts.require("LoanOpenings");
+const LoanClosingsEvents = artifacts.require("LoanClosingsEvents");
```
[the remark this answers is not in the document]
cwsnt: Removed in this [commit](https://github.com/DistributedCollective/Sovryn-smart-contracts/pull/453/commits/9174699f93acf8c4458d4c552b1180f82fb45214)

contracts/modules/LoanClosingsShared.sol
```diff
@@ -37,6 +38,25 @@ contract LoanClosingsShared is
...
+        require(loanTokenPoolAddress != address(0), "Invalid loan token pool address");
+
+        uint256 previousITokenSupply = ILoanTokenModules(loanTokenPoolAddress).totalSupply();
+
+        _;
+
+        /// Validate iToken total supply
+        require(
+            previousITokenSupply == ILoanTokenModules(loanTokenPoolAddress).totalSupply(),
+            "mismatch loan token supply"
```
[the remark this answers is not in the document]
cwsnt: Updated in this [commit](https://github.com/DistributedCollective/Sovryn-smart-contracts/pull/453/commits/9174699f93acf8c4458d4c552b1180f82fb45214)

contracts/testhelpers/TestDiscRBTC.sol
cwsnt: Updated in this [commit](https://github.com/DistributedCollective/Sovryn-smart-contracts/pull/453/commits/9174699f93acf8c4458d4c552b1180f82fb45214)

contracts/modules/LoanClosingsShared.sol
```diff
@@ -37,6 +38,25 @@ contract LoanClosingsShared is
 
     enum CloseTypes { Deposit, Swap, Liquidation }
 
+    /** modifier for invariant check */
+    modifier iTokenSupplyUnchanged(bytes32 loanId) {
+        Loan storage loanLocal = loans[loanId];
+        LoanParams storage loanParamsLocal = loanParams[loanLocal.loanParamsId];
+
+        address loanTokenPoolAddress = underlyingToLoanPool[loanParamsLocal.loanToken];
```
ororopickpocket, line 46: is this necessary in 2 steps? can't we just read the lender from the loanLocal object?

contracts/testhelpers/TestCrossReentrancyRBTC.sol
```diff
@@ -0,0 +1,132 @@
...
+            _borrower,
+            _receiver,
+            ""
+        );
+
+        // step2 mint 15 iWRBTC
+        // prerequisite: this contract should have WRBTC
+        uint256 _WRBTCBalance = IERC20(WRBTC).balanceOf(address(this));
+        IERC20(WRBTC).approve(loanTokenWRBTC, _WRBTCBalance);
+        ILoanTokenModules(loanTokenWRBTC).mint(_receiver, 15 ether);
```
ororopickpocket, line 96: why mint here?

contracts/testhelpers/TestCrossReentrancyERC777.sol
```diff
@@ -0,0 +1,152 @@
...
+            _borrower,
+            _receiver,
+            ""
+        );
+
+        // step2 mint iSUSD
+        // prerequisite: this contract should have SUSD
+        uint256 _SUSDBalance = IERC20(SUSD).balanceOf(address(this));
+        IERC20(SUSD).approve(loanToken, _SUSDBalance);
+        ILoanTokenModules(loanToken).mint(_receiver, 100 ether);
```
ororopickpocket, line 103: why mint here?

contracts/modules/LoanClosingsShared.sol
cwsnt: good point, just realized there is lender property in the struct.
Updated

contracts/modules/LoanClosingsShared.sol
```diff
@@ -37,6 +38,24 @@ contract LoanClosingsShared is
 
     enum CloseTypes { Deposit, Swap, Liquidation }
 
+    /** modifier for invariant check */
+    modifier iTokenSupplyUnchanged(bytes32 loanId) {
+        Loan storage loanLocal = loans[loanId];
+        LoanParams storage loanParamsLocal = loanParams[loanLocal.loanParamsId];
```
ororopickpocket, line 44: loanParamsLocal are not required here anymore, no?

ororopickpocket (changes requested):

contracts/testhelpers/TestCrossReentrancyERC777.sol
cwsnt: removed the mint

contracts/testhelpers/TestCrossReentrancyRBTC.sol
cwsnt: removed the mint

contracts/modules/LoanClosingsShared.sol
cwsnt: removed

tests/Utils/initializer.js
```diff
@@ -170,7 +178,7 @@ const getLoanTokenLogic = async (isMockLoanToken = false) => {
 
     return [loanTokenLogicProxy, loanTokenLogicBeacon];
 };
-const getLoanTokenLogicWrbtc = async () => {
+const getLoanTokenLogicWrbtc = async (isMockLoanToken = false) => {
```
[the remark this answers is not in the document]
cwsnt: As per discussion, this fix will be addressed in the other PR

tests/loan-token/BorrowingTestToken.test.js
```diff
@@ -49,22 +57,25 @@ const hunEth = new BN(wei("100", "ether"));
...
         WRBTC = await getWRBTC();
         BZRX = await getBZRX();
         priceFeeds = await getPriceFeeds(WRBTC, SUSD, RBTC, BZRX);
 
         sovryn = await getSovryn(WRBTC, SUSD, RBTC, priceFeeds);
 
-        loanToken = await getLoanToken(owner, sovryn, WRBTC, SUSD);
-        loanTokenWRBTC = await getLoanTokenWRBTC(owner, sovryn, WRBTC, SUSD);
+        loanToken = await getLoanToken(owner, sovryn, WRBTC, SUSD, true);
+        loanTokenWRBTC = await getLoanTokenWRBTC(owner, sovryn, WRBTC, SUSD, true);
```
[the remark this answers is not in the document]
cwsnt: As per discussion, this fix will be addressed in the other PR

contracts/farm/LiquidityMining.sol
```diff
@@ -628,8 +628,6 @@ contract LiquidityMining is ILiquidityMining, LiquidityMiningStorage {
         user.accumulatedReward = 0;
         pool.poolToken.safeTransfer(address(msg.sender), userAmount);
 
-        _updateRewardDebt(pool, user);
```
tjcloa, line 631: it is removed as all the params are set directly (nullified) above

tests/affiliates/affiliates_integration.test.js
```diff
@@ -81,6 +82,8 @@ contract("Affiliates", (accounts) => {
     let referrerFee;
 
     async function deploymentAndInitFixture(_wallets, _provider) {
+        // need to deploy the mutex in the initialization. Otherwise, the global reentrancy prevetion will not be working & throw an error
```
tjcloa, line 85:
```suggestion  
        // Need to deploy the mutex in the initialization. Otherwise, the global reentrancy prevention will not be working & throw an error.
```

tests/loan-token/Administration.test.js
```diff
@@ -50,6 +51,9 @@ contract("LoanTokenAdministration", (accounts) => {
     let sovryn, SUSD, WRBTC, RBTC, BZRX, loanToken, loanTokenWRBTC;
 
     async function deploymentAndInitFixture(_wallets, _provider) {
+        // need to deploy the mutex in the initialization. Otherwise, the global reentrancy prevetion will not be working & throw an error
```
tjcloa, line 54:
```suggestion  
        // Need to deploy the mutex in the initialization. Otherwise, the global reentrancy prevention will not be working & throw an error.
```

tests/loan-token/BorrowingTestToken.test.js
```diff
@@ -49,22 +56,26 @@ const hunEth = new BN(wei("100", "ether"));
...
         await loan_pool_setup(sovryn, owner, RBTC, WRBTC, SUSD, loanToken, loanTokenWRBTC);
 
         SOV = await getSOV(sovryn, priceFeeds, SUSD, accounts);
+
+        // need to deploy the mutex in the initialization. Otherwise, the global reentrancy prevetion will not be working & throw an error
```
tjcloa, line 77:
```suggestion  
        // Need to deploy the mutex in the initialization. Otherwise, the global reentrancy prevention will not be working & throw an error.
```

tests/protocol/LiquidationTestToken.test.js
```diff
@@ -52,6 +53,9 @@ contract("ProtocolLiquidationTestToken", (accounts) => {
     let sovryn, SUSD, WRBTC, RBTC, BZRX, loanToken, loanTokenWRBTC, priceFeeds, SOV;
 
     async function deploymentAndInitFixture(_wallets, _provider) {
+        // need to deploy the mutex in the initialization. Otherwise, the global reentrancy prevetion will not be working & throw an error
```
tjcloa, line 56:
```suggestion  
        // Need to deploy the mutex in the initialization. Otherwise, the global reentrancy prevention will not be working & throw an error.
```

tjcloa (approved):
but for cosmetic comments improvements proposed in comments, 
lgtm

jjmr007:
Hey guys, great work!... just a cosmetic suggestion:

On the "Nature of the Exploit" explanation, the 3rd paragraph in the first section "Background information" it says:

"**The share which is currently borrowed out is held by the protocol,**"

Which is not 100% accurate (remember the borrowings are actually taken by the borrower, and only the funds borrowed for margin trades are held by the protocol). So would be nice to change to:

"**The accountant register of those funds currently borrowed out, is held by the protocol,**"

ororopickpocket (approved):

PR #467, opened 2022-12-09
Exhaustive staking contract tests and some contract changes

koirikivi:
Ready for review/merging


tests/staking/StakingTest.js
```diff
@@ -72,6 +74,8 @@ contract("Staking", (accounts) => {
         await staking.setVestingRegistry(vesting.address);
 
         MAX_VOTING_WEIGHT = await staking.getStorageMaxVotingWeight.call();
+        MAX_DURATION = await staking.MAX_DURATION();
```
tjcloa, line 77:
```suggestion 
        MAX_DURATION = await staking.getStorageMaxDurationToStakeTokens();
```

tests/staking/StakingTest.js
```diff
@@ -1595,7 +1587,165 @@ contract("Staking", (accounts) => {
...
+                date,
+                startDate,
+                blockNumber
+            );
+            expect(weightedStake).to.be.bignumber.eq(
+                getAmountWithWeight("10000", date, startDate)
+            );
+        });
+
+        xit("if date is not a valid lock date, the function will return the weighted stake at the closest lock date prior to date", async () => {
```
tjcloa, line 1691: the spec of SOV-1216 says 
`if date is not a valid lock date, the function will return the weighted stake at the closest lock date AFTER date`

contracts/governance/Staking/modules/StakingWithdrawModule.sol
```diff
@@ -250,7 +251,10 @@ contract StakingWithdrawModule is IFunctionsList, StakingShared, CheckpointsShar
         bool isGovernance
     ) internal {
         if (_isVestingContract(msg.sender)) {
-            uint256 nextLock = until.add(TWO_WEEKS);
+            // adjust nextLock to the next valid lock date to make sure we don't accidentally
+            // withdraw stakes that are in the future and would get slashed (if until is not
+            // a valid lock date)
+            uint256 nextLock = _adjustDateForOrigin(until.add(TWO_WEEKS));
```
ororopickpocket, line 257: this is correct, but we could save some gas by adjusting `until` in the public function `withdraw` , so we do not have to adjust 2 dates but just 1 (if until is already adjusted, we know that until + 2 weeks is valid.

PR #480, opened 2023-01-25
Feature/update cancel team vesting

cwsnt:


ororopickpocket:
@cwsnt looks good in general, but could you please just add comments with the test case description from here: https://sovryn.atlassian.net/jira/software/c/projects/SOV/boards/4?modal=detail&selectedIssue=SOV-1236 to the places in the code where these test cases are being checked? this would make the review much easier.

contracts/governance/Staking/modules/StakingWithdrawModule.sol
```diff
@@ -245,12 +245,9 @@ contract StakingWithdrawModule is IFunctionsList, StakingShared, CheckpointsShar
     ) internal {
         address vesting = vestingConfig.vestingAddress;
 
-        until = _adjustDateForOrigin(until);
+        until = _timestampToLockDate(until);
         _validateWithdrawParams(vesting, amount, until);
 
-        /// @dev Determine the receiver.
-        if (receiver == address(0)) receiver = msg.sender;
```
jameshowlett977, line 252: I think it's very unlikely that receiver is zero for this function, but it could be the case, so we need this check.

jameshowlett977 (changes requested):
- the function is executable if the contract is paused

Please add modifier **whenNotPaused**.

contracts/governance/Staking/modules/StakingWithdrawModule.sol
cwsnt: Currently the `_withdrawFromTeamVesting` is internal function which will only be execute by `_cancelTeamVesting` function, and there is a check `require(_receiver != address(0), "receiver address invalid");` that will prevent the receiver being 0 address.
So could we agree here that we don't need this check anymore?

jameshowlett977: Yes, thank you for the clarification.

jameshowlett977 (approved):

ororopickpocket (approved):

PR #489, opened 2023-02-24
feeSharingProxy:remove sender check in fallback

cwsnt:


ororopickpocket (approved):
I'm approving this now to not block the FE team and delay the D2 launch, however, I would like you to check if the problem could also be solves by allowing the sender being the iRBTC contract.

tjcloa (approved):

PR #491, opened 2023-03-03
script for creating SIP 56

ororopickpocket:


tjcloa (commented):
not sure abt implications of using `GovernorOwner` for non-executable proposals instead of `GovernorAdmin` 
apart from this lgtm

PR #492, opened 2023-03-08
fix: enable set vesting stakes if lockedDate lesser than blockTimestamp

cwsnt:


tjcloa (approved):
pls fix the deprecation warning comment (see inline comments) and a typo

PR #496, opened 2023-04-13
SOV-2181 fix FeeSharingCollector user fees withdraw

tjcloa:
The issue if a user has 0 stake at the time of the first _maxCheckpoints when calling `FeeSharingCollector.withdraw(...)` function, it  leads to the function reverting preventing users from claiming their fees share.


contracts/governance/FeeSharingCollector/FeeSharingCollector.sol
```diff
@@ -288,38 +280,125 @@ contract FeeSharingCollector is
...
+     * */
+    function withdrawStartingFromCheckpoint(
+        address _token,
+        uint256 _fromCheckpoint,
+        uint32 _maxCheckpoints,
+        address _receiver
+    ) public validFromCheckpointParam(_fromCheckpoint, msg.sender, _token) nonReentrant {
+        if (_fromCheckpoint > processedCheckpoints[msg.sender][_token]) {
+            processedCheckpoints[msg.sender][_token] = _fromCheckpoint.sub(1);
+        }
```
ororopickpocket, line 397: nice way to solve it!

ororopickpocket (approved):

contracts/governance/FeeSharingCollector/FeeSharingCollectorStorage.sol
```diff
@@ -9,30 +9,25 @@ import "../../mixins/EnumerableAddressSet.sol";
...
-
     IProtocol public protocol;
     IStaking public staking;
 
     /// @notice Checkpoints by index per pool token address
     mapping(address => mapping(uint256 => Checkpoint)) public tokenCheckpoints;
 
     /// @notice The number of checkpoints for each pool token address.
-    mapping(address => uint256) public numTokenCheckpoints;
+    mapping(address => uint256) public totalTokenCheckpoints;
```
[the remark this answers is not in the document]
cwsnt: This might cause the backward compatibility issue?

tests/FeeSharingCollectorTest.js
```diff
@@ -291,21 +288,508 @@ contract("FeeSharingCollectorProxy:", (accounts) => {
...
+    describe("withdrawStartingFromCheckpoint, withdrawRBTCStartingFromCheckpoint and getNextPositiveUserCheckpoint", () => {
+        let snapshot;
+        before(async () => {
+            await loadFixture(protocolDeploymentFixture);
+        });
+        beforeEach(async () => {
+            snapshot = await takeSnapshot();
+        });
+        afterEach(async () => {
+            await snapshot.restore();
```
cwsnt, line 308: Good idea to introduce the snapshot here.
Might be useful for other test cases

tests/FeeSharingCollectorTest.js
```diff
@@ -291,21 +288,508 @@ contract("FeeSharingCollectorProxy:", (accounts) => {
...
+                .add(borrowingFeeTokensHeld);
+            let feeAmount = await setFeeTokensHeld(
+                lendingFeeTokensHeld,
+                tradingFeeTokensHeld,
+                borrowingFeeTokensHeld,
+                false,
+                true
+            );
+
+            tx = await feeSharingCollector.withdrawFees([SOVToken.address]);
```
cwsnt, line 402: No need to do assignment here (tx = ).
And seems like it's missing another check after .withdrawFeess()

tests/FeeSharingCollectorTest.js
```diff
@@ -291,21 +288,508 @@ contract("FeeSharingCollectorProxy:", (accounts) => {
...
+            expect([
+                nextCheckpoint.checkpointNum.toNumber(),
+                nextCheckpoint.hasSkippedCheckpoints,
+                nextCheckpoint.hasFees,
+            ]).to.eql([10, true, true]);
+        });
+    });
 
     describe("FeeSharingCollectorProxy", () => {
+        /*before(async () => {
```
cwsnt, line 779: Why is this commented out?

contracts/governance/FeeSharingCollector/FeeSharingCollectorStorage.sol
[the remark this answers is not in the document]
tjcloa: good catch  
it does break backwards compatibility    
my reasoning to rename was driven by this array index vs number of elements values which causes confusion and `totalTokenCheckpoints` looked much more clear and unambiguous than `numTotalCheckpoints` (is it an index or a total **number** of checkpoints?).  
to preserve backwards compatibility, i'm adding a getter function with potential deprecation with the respective comment.

tests/FeeSharingCollectorTest.js
tjcloa: this is a workaround in this particular case 
`loadFixture` takes a snapshot and should revert to it when subsequently called  
but sometimes (and in this particular case too) it doesn't revert or not completely reverts for unclear reason causing the upfollowing tests to fail - they pass when run isolated though

tests/FeeSharingCollectorTest.js
tjcloa: a leftover after experiments with restoring snapshots  
not necessary here as restored in the upfollowing tests as needed but uncommented to keep it uniform

tests/FeeSharingCollectorTest.js
tjcloa: a copypaste lefover    
it is covered in the previous test  
removing it

tjcloa: a copypaste leftover  
removing it

contracts/governance/FeeSharingCollector/FeeSharingCollectorStorage.sol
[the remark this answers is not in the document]
cwsnt: Should we re-introduce the `numTotalCheckpoints` getter back to the unit test?

cwsnt:
Approved this PR, but just left with a small comment
https://github.com/DistributedCollective/Sovryn-smart-contracts/pull/496#discussion_r1173762959

cwsnt (approved):

contracts/governance/FeeSharingCollector/FeeSharingCollectorStorage.sol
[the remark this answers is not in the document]
tjcloa: added [here](https://github.com/DistributedCollective/Sovryn-smart-contracts/commit/8af7c2494693341466abc3127f0aca858698f881)

ororopickpocket (approved):

ororopickpocket (approved):

cwsnt (approved):

cwsnt (commented):
There is a conflict, need to solve it

tjcloa:
all conflicts resolved

cwsnt (approved):
lgtm

contracts/governance/FeeSharingCollector/FeeSharingCollector.sol
```diff
@@ -373,76 +457,198 @@ contract FeeSharingCollector is
...
+        for (uint256 i = startFrom; i < end; i++) {
+            Checkpoint storage tokenCheckpoint = tokenCheckpoints[_token][i];
+            uint96 weightedStake =
+                staking.getPriorWeightedStake(
+                    _user,
+                    tokenCheckpoint.blockNumber - 1,
+                    tokenCheckpoint.timestamp
+                );
+            if (weightedStake > 0) {
+                // i is the index and we need to return checkpoint num which is i + 1
```
ororopickpocket, line 569: would it be simpler to just return the index? then, we wouldn't need to subtract 1 from the passed param on the withdraw functions, no?

ororopickpocket (approved):

contracts/governance/FeeSharingCollector/FeeSharingCollector.sol
tjcloa: by design most of the functions return checkpoints number rather then index. it looks weird at first but it makes sense because we mostly need to work the next checkpoint, which is the processedCheckpoints[user][token] - index and number of processed tokens at the same time. therefore we either just use the returned values for most of the cases as index for the next checkpoint or add 1 every time. both are fine, i'm just sticking with the existing solution.

PR #498, opened 2023-04-26
SIP script for stability pool subsidies

cwsnt:


ororopickpocket (commented):
Looks good so far. Is the APR already set to 5% at this point? and the reward manager to exchequer? is there a script for it?

cwsnt:
> Looks good so far. Is the APR already set to 5% at this point? and the reward manager to exchequer? is there a script for it?

The APR will be set at the contract initialization of community issuance, as well as the reward manager address right after it.

ororopickpocket (approved):

tjcloa (changes requested):
pls use hh sip scripts
otherwise it is pretty hard to do QA testing

cwsnt:
> pls use hh sip scripts otherwise it is pretty hard to do QA testing

Added sip scripts for HH

PR #499, opened 2023-05-24
SOV-2197 Recover incorrect fee allocation on feeSharingCollector

cwsnt:


tjcloa:
Missing a test for revert sending RBTC which could be the case when the owner is a contract: "FeeSharingCollector::recoverIncorrectAllocatedFees: Withdrawal rbtc failed"

tjcloa (commented):
Missing a test for revert sending RBTC which could be the case if the receiver is a contract: "FeeSharingCollector::recoverIncorrectAllocatedFees: Withdrawal rbtc failed"

ororopickpocket (approved):
in general it would be better to read the owner only once and store it locally to save gas, but because this is a one-time function it doesn't really matter

tjcloa (approved):

PR #509, opened 2023-07-21
SOV-2672 feat: add pauser role in protocol

cwsnt:


contracts/modules/ProtocolSettings.sol
```diff
@@ -393,6 +395,30 @@ contract ProtocolSettings is
...
     }
 
+    /**
+     * @notice Set the pauser address of sovryn protocol.
+     *
+     * only pauser or owner can perform this action.
+     *
+     * @param newPauser The new address of the pauser.
+     * */
+    function setPauser(address newPauser) external onlyPauserOrOwner {
```
tjcloa, line 405: it should be exclusively the owner's privilege to set a pauser

tjcloa (changes requested):

tests/other/ProtocolSettings.test.js
```diff
@@ -968,6 +968,40 @@ contract("ProtocolSettings", (accounts) => {
             expect(event["newRolloverFlexFeePercent"] == new_percent.toString()).to.be.true;
             expect((await sovryn.rolloverFlexFeePercent()).eq(new_percent)).to.be.true;
         });
+
+        it("Test set pauser", async () => {
```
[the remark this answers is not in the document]
cwsnt: updated

contracts/core/State.sol
```diff
@@ -228,4 +230,9 @@ contract State is Objects, ReentrancyGuard, SharedReentrancyGuard, Ownable {
             logicTargetsSet.removeBytes32(bytes32(sig));
         }
     }
+
+    modifier onlyPauserOrOwner() {
```
tjcloa, line 234: pls move the modifier to where it is actually used for clarity - ProtocolSettings

tests/other/ProtocolSettings.test.js
```diff
@@ -968,6 +968,65 @@ contract("ProtocolSettings", (accounts) => {
...
+            const data2 = sovryn.contract.methods.setPauser(pauser2).encodeABI();
+            const tx2 = await multisig.submitTransaction(dest, val, data2, { from: accounts[0] });
+            const txId2 = tx2.logs.filter((item) => item.event == "Submission")[0].args[
+                "transactionId"
+            ];
+            await multisig.confirmTransaction(txId2, { from: accounts[1] });
+            expect((await sovryn.getPauser()) == pauser2).to.be.true;
+        });
+
+        it("Should be able to set pauser to 0 address", async () => {
```
tjcloa, line 1016: setting to and from(!) 0 address - the test should prove ability to change the address from non-zero to zero and the other way round

tjcloa (changes requested):

contracts/core/State.sol
cwsnt: i'd prefer to keep it in the state contract, so that it can be reusable by other module

tests/other/ProtocolSettings.test.js
cwsnt: updated

contracts/core/State.sol
tjcloa: ok. makes sense.

PR #510, opened 2023-07-24
SOV-2680 feat:add pauser role in loantoken

cwsnt:


ororopickpocket (approved):

tjcloa (changes requested):
pls add hh deployment and upgrade scripts of the beacons to stick with the hh deploy format we are using and that will be used for integrations both internal and external

deployment/deploy/2030-deploy-ReplaceLoanTokenBeacon.js
```diff
@@ -0,0 +1,409 @@
...
+                )
+            );
+        }
+
+        /** 5. Replace loan token logic beacon in proxy */
+        loanTokenWithProxyABI = Contract.from_abi(
+            "loanTokenWithProxyABI",
+            (address = loanTokenAddress),
+            (abi = LoanTokenLogicProxy.abi),
+            (owner = conf.acct)
```
tjcloa, line 229: looks like AI forgot to transpile this to js ;)

tjcloa: pls make sure the scripts actually work - run it on forked chains

deployment/deploy/2030-deploy-ReplaceLoanTokenBeacon.js
```diff
@@ -0,0 +1,409 @@
...
+        );
+        let data = loanTokenLogicBeaconLMInterface.encodeFunctionData("registerLoanTokenModule", [
+            loanTokenLogicLMDeployment.address,
+        ]);
+
+        /** iBPRO */
+        data = loanTokenLogicProxyInterface.encodeFunctionData("setBeaconAddress", [
```
tjcloa, line 236: pls wrap it into an array/map to make the code consize - same for similar cases in the script

deployment/deployments/rskSovrynMainnet/LoanTokenLogicProxy.json
```diff
@@ -0,0 +1,636 @@
+{
+  "abi": [
```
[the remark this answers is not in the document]
tjcloa: 1) it looks like there is no way to tell what LoanTokenLogicProxy contract address is used in the tokens,   
pls add a view function getTarget() to the LoanTokenLogicProxy contract to return `_target`
1) python deployment script deploy a separate LoanTokenLogicProxy for each LoanToken while one contract can be used 
pls add deployment script for the LoanTokenLogicProxy and register the same instance in all the tokens - use arrays/maps and cycle through

deployment/deploy/2030-deploy-ReplaceLoanTokenBeacon.js
```diff
@@ -0,0 +1,409 @@
...
+            loanTokenLogicProxyDeployment.abi,
+            loanTokenDeploymentIRBTC.address
+        );
+        await loanTokenLogicContractIRBTC.setBeaconAddress(
+            loanTokenLogicBeaconWRBTCDeployment.address
+        );
+    }
+};
+func.tags = ["DeployLoanTokenBeacon"];
+func.dependencies = ["DeployLoanTokenBeacon"];
```
tjcloa, line 408: tags and dependency cannot be the same

tjcloa (changes requested):

deployment/deploy/2030-deploy-ReplaceLoanTokenBeacon.js
cwsnt: Updated & tested works on the forked chain

deployment/deploy/2030-deploy-ReplaceLoanTokenBeacon.js
cwsnt: updated

deployment/deploy/2030-deploy-ReplaceLoanTokenBeacon.js
cwsnt: Updated

deployment/deployments/rskSovrynTestnet/LoanTokenLogicProxy.json
```diff
@@ -0,0 +1,636 @@
+{
+  "abi": [
```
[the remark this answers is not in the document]
cwsnt: Removed this since it's not being used

deployment/deployments/rskSovrynMainnet/LoanTokenLogicProxy.json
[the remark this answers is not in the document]
cwsnt: Updated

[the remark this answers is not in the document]
tjcloa: right when speaking about deployment and using current LoanTokenLogicProxy implementation - all you need is interface  
but what i mean - we do need it in case we want to replace LoanTokenLogicProxy and now we have it deployed per contract which is suboptimal and hard to manage and for clarity we do need to tell what LoanTokenLogicProxy contract is used as a proxy for LoanTokens (iTokens) - it should be deployed and verified and deployment and replacement script should be ready and only one contract should be used for all the tokens for now

tjcloa (changes requested):
almost there  
but we do need to fix this LoanTokenLogicIssue

deployment/deployments/rskSovrynMainnet/LoanTokenLogicProxy.json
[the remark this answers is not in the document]
tjcloa: it was decided that this issue to to be done in a separate ticket SOV-2857

tjcloa (approved):
approving the PR taking into account that the fix this LoanTokenLogicProxy issue to be done in a separate ticket SOV-2857

PR #514, opened 2023-08-17
SOV-585 transfer the contract guardianship

ororopickpocket:
SIP-0047 and transfer guardianship to the Contracts Guardian Safe multisig


scripts/sip/sip_interaction.py
```diff
@@ -565,3 +567,21 @@ def createProposalSIP0065():
...
+    stakingLogic = Contract.from_abi("StakingLogic5", address=contracts['StakingLogic5'], abi=interface.IStaking.abi, owner=acct)
+
+    # Action
+    targets = [contracts['Staking']]
+    values = [0, 0]
+    signatures = ["addPauser(address)", "removePauser(address)"]
+    data1 = stakingLogic.addPauser.encode_input(newGuardian)
+    data2 = stakingLogic.removePauser.encode_input(contracts['multisig'])
+    datas = ["0x" + data1[10:], "0x" + data2[10:]]
+    description = "SIP-0042: Staking Security Update, Details: https://github.com/DistributedCollective/SIPS/blob/7c1a44b/SIP-0042.md, sha256: 522e1e65c49ec028d81c3a1f94a47354c2f6287c2d90c6eec8f06dcc17a1ebcc"
```
cwsnt, line 581: Probably need to add @todo mark here, since we need to update the description as well later?

cwsnt (commented):
We actually have the sip script in the hardhat task, i think we should use hardhat task instead
https://github.com/DistributedCollective/Sovryn-smart-contracts/blob/408388cfa1c073cd52586d4289978971c49ca02b/hardhat/tasks/sips.js#L51

scripts/sip/sip_interaction.py
ororopickpocket: you're right. done

ororopickpocket:
> We actually have the sip script in the hardhat task, i think we should use hardhat task instead

Going forward you're probably right. I won't change it in this PR though. It will be yours and Tyrones choice to either go ahead with the prepared sip creation script ot transcribe it to hardhat

cwsnt (approved):

PR #528, opened 2023-11-05
Sov 3274 airswap proxy

dharkmattr:


contracts/airswap/AirSwapFeeConnector.sol
```diff
@@ -0,0 +1,202 @@
...
+        address senderToken;
+        uint256 totalSenderAmount;
+        uint8 v;
+        bytes32 r;
+        bytes32 s;
+    }
+
+    uint256 public constant POINTS = 1000;
+
+    uint256 public inputFeeInPoints = 0;
```
cwsnt, line 30: Probably can just removed the 0 initialization here, since it will use the zero value by default

contracts/airswap/AirSwapFeeConnector.sol
```diff
@@ -0,0 +1,202 @@
...
+    uint256 public outputFeeInPoints = 0;
```
cwsnt, line 31: Probably can just removed the 0 initialization here, since it will use the zero value by default

contracts/airswap/AirSwapFeeConnector.sol
```diff
@@ -0,0 +1,202 @@
...
+
+    address public feeVaultAddress = address(0);
```
cwsnt, line 33: Probably can just removed the 0 initialization here, since it will use the zero value by default

contracts/airswap/AirSwapFeeConnector.sol
```diff
@@ -0,0 +1,202 @@
...
+    address public swapERC20Address = address(0);
```
cwsnt, line 34: Probably can just removed the 0 initialization here, since it will use the zero value by default

contracts/airswap/AirSwapFeeConnector.sol
```diff
@@ -0,0 +1,202 @@
...
+        uint256 expiry;
+        address signerWallet;
+        address signerToken;
+        uint256 signerAmount;
+        address senderToken;
+        uint256 totalSenderAmount;
+        uint8 v;
+        bytes32 r;
+        bytes32 s;
+    }
```
cwsnt, line 26: Can add the constructor for the initialization value of the configs

contracts/airswap/AirSwapFeeConnector.sol
```diff
@@ -0,0 +1,202 @@
...
+                swapRequest.sender,
+                address(this),
+                swapRequest.totalSenderAmount
+            ),
+            "transfer failed 1"
+        );
+
+        // then we collect the input fee
+        uint256 inputFee = calculateInputFee(swapRequest.totalSenderAmount);
+        require(
```
cwsnt, line 150: The transfer here can be combined with the output fee.
So will save some gas here

contracts/airswap/AirSwapFeeConnector.sol
```diff
@@ -0,0 +1,202 @@
...
+        address signerToken;
+        uint256 signerAmount;
+        address senderToken;
+        uint256 totalSenderAmount;
+        uint8 v;
+        bytes32 r;
+        bytes32 s;
+    }
+
+    uint256 public constant POINTS = 1000;
```
cwsnt, line 28: By having the points = 1000, this means the smallest fee that we will have is 0.1%, am i correct?

cwsnt (changes requested):
Need to add the deployment script.
Can refer to this https://github.com/DistributedCollective/Sovryn-smart-contracts/tree/development/deployment/deploy

PR #534, opened 2024-01-24
SOV-3686 introduce admin manager role

cwsnt:


contracts/utils/AdminRole.sol
```diff
@@ -1,8 +1,9 @@
 pragma solidity 0.5.17;
 
 import "../openzeppelin/Ownable.sol";
+import "./AdminManagerRole.sol";
 
-contract AdminRole is Ownable {
+contract AdminRole is Ownable, AdminManagerRole {
```
tjcloa, line 6: we need AdminManager only for VestingRegistry which is an exception (because admins are EOAs creating vesting contracts to avoid huge amount of multisig txs to sign on each vesting creation) and can be set and removed following current needs, but in all other cases the admin role is assigned to a multisig and should be managed by Bitocracy only. therefore there is no justification on why LiquidityMining should have this admin manager role too.  
to keep it consistent with the ticket requirements, i suggest to introduce a new role AdminRoleManaged rather than extending the existing AdminRole.  
this way we have separation of concerns and clear alternative should we ever need to use it again.

tjcloa (changes requested):
see inline comments

koirikivi (approved):

deployment/deploy/3000-deploy-VestingRegistry.js
```diff
@@ -0,0 +1,25 @@
...
+        ethers,
+    } = hre;
+    const { deployer } = await getNamedAccounts();
+    log(col.bgYellow("Deploying VestingRegistry..."));
+    await deployWithCustomProxy(
+        deployer,
+        "VestingRegistryLogic",
+        "VestingRegistryProxy",
+        "",
+        "VestingRegistry_Proxy",
```
tjcloa, line 20: because there is already VestingRegistryProxy deployment data saved, either rename the deployment instance:  
```suggestion
        "VestingRegistryProxy",
```
or rename the deployment file on both mainnet and testnet

deployment/deploy/3000-deploy-VestingRegistry.js
```diff
@@ -0,0 +1,25 @@
...
+        true
+    );
```
tjcloa: There is a confusion because we have VestingRegistry contract which was merged into a single vesting registry together with VestingRegistry2 and VestignRegistry3 - all of those are deprecated.    
I suggest the following clean-up for clarity:
1) remove legacy vesting registry contracts VestingRegistry, VestingRegistry2, VestingRegistry3
2) rename VestingRegistryLogic to VestingRegistry
3) update tests and deployment naming  
current implementation is not consistent with sipArgs.   
suggested change:  
```suggestion
    await deployWithCustomProxy(
        deployer,
        "VestingRegistry",
        "VestingRegistryProxy",
        "",
        "VestingRegistry_Proxy",
        true
    );
```

hardhat/tasks/adminManagerRole.js
```diff
@@ -0,0 +1,33 @@
+const { task } = require("hardhat/config");
+const Logs = require("node-logs");
+const logger = new Logs().showInConsole(true);
+const { sendWithMultisig } = require("../../deployment/helpers/helpers");
+
+task("setAdminManager", "SetAdminManager to the contract that implement adminManagerRole")
+    .addOptionalParam("signer", "Signer name: 'signer' or 'deployer'", "deployer")
+    .addParam("adminManagerTarget", "Admin manager target: 'MultiSigWallet'")
```
tjcloa:
```suggestion
    .addParam("adminManager", "New admin manager: 'MultiSigWallet'")
```

hardhat/tasks/adminManagerRole.js
```diff
@@ -0,0 +1,33 @@
...
+    .addParam("contractTarget", "Contract name target: e.g: 'VestingRegistry'")
```
tjcloa: pls add support for address input - see multisig hh tasks for reference  
```suggestion
    .addParam("contract", "Contract name or address to set admin manager: e.g: 'VestingRegistry'")
```

deployment/deploy/3000-deploy-VestingRegistry.js
```diff
@@ -0,0 +1,25 @@
+const path = require("path");
+const { getContractNameFromScriptFileName } = require("../helpers/utils");
+const { sendWithMultisig } = require("../helpers/helpers");
+const { deployWithCustomProxy } = require("../helpers/helpers");
+const col = require("cli-color");
+//const deploymentName = getContractNameFromScriptFileName(path.basename(__filename));
+const func = async function (hre) {
```
tjcloa, line 7: missing setting admin manager role on deployment/upgrade - should be included into the contract's initialization

contracts/utils/AdminManagerRole.sol
```diff
@@ -0,0 +1,62 @@
...
+        require(isOwner() || msg.sender == getAdminManager(), "unauthorized");
+        _;
+    }
+
+    /**
+     * @notice Set new admin manager.
+     * @param _newAdminManager The addresses of the account to grant permissions.
+     * */
+    function setAdminManager(address _newAdminManager) public onlyOwner {
+        require(_newAdminManager != address(0), "invalid admin manager");
```
tjcloa:
```suggestion
        require(_newAdminManager != address(0), "Use removeAdminManager function to set adminManager to 0");
```

tjcloa (changes requested):
pls see inline comments + resolve conflicts

tjcloa (approved):

PR #536, opened 2024-02-01
SOV-3763 StakingRewardsOS

tjcloa:


contracts/governance/StakingRewards/StakingRewardsOs.sol
```diff
@@ -0,0 +1,256 @@
...
+        if (stopBlock > 0 && stopBlock < _block) {
+            uint256 previousWeightedStake = staking.getPriorWeightedStake(
+                _staker,
+                stopBlock,
+                _date
+            );
+            if (previousWeightedStake < weightedStake) {
+                weightedStake = previousWeightedStake;
+            }
+        }
```
ororopickpocket: This logic should not be necessary anymore. The existing SOV staking rewards contract continues after the stopBlock, but increases of the stake are no longer considered. This osSOV staking rewards contract truly stops after the stop block, so this logic here is no longer required. I would just return 0 in case that (stopBlock > 0 &&  stopBlock < _block).

contracts/governance/StakingRewards/StakingRewardsOs.sol
```diff
@@ -0,0 +1,256 @@
...
+     * @dev Base rate is annual, but we pay interest for 14 days,
+     * which is 1/26 of one staking year (1092 days)
+     * @param _staker Staker address
+     * @param _amount the reward amount
+     * */
+    function _payReward(address _staker, uint256 _amount) internal {
+        require(_amount != 0, "amount invalid");
+        claimedBalances[_staker] = claimedBalances[_staker].add(_amount);
+        osSOV.mint(_staker, _amount);
+        emit RewardPaid(_staker, _amount);
```
ororopickpocket: Renaming this event from RewardWithdrawn to RewardPaid creates unnecessary work for the CoL (needs to be added to the graph and the logic of the frontend needs to be adapted as well - instead of just using what's already built).  I would therefore advise to keep the old event name.

ororopickpocket: You are also changing a number of variable names throughout the code. I would not change the name of public variables to have a uniform ABI for both staking rewards contracts. Or at least not change whichever parts of the ABI the CoL is using.

contracts/governance/StakingRewards/StakingRewardsOs.sol
```diff
@@ -0,0 +1,256 @@
...
+     * @param _startTime The time from which the staking rewards calculation shall restart.
+     * @return The timestamp of last withdrawal
+     * @return The accumulated reward
+     */
+    function getStakerCurrentReward(
+        bool _considerMaxDuration,
+        uint256 _startTime
+    ) external view returns (uint256 nextWithdrawTimestamp, uint256 amount) {
+        return _getStakerCurrentReward(msg.sender, _considerMaxDuration, _startTime);
+    }
```
ororopickpocket, line 190: I like that you split this function into an external and internal one, passing msg.sender as the _staker address to the internal function. I would add another external function to retrieve the current reward for a given address, because this is currently a very annoying shortcoming of the existing staking rewards contract.

contracts/governance/StakingRewards/StakingRewardsOs.sol
```diff
@@ -0,0 +1,256 @@
...
+                ? staking.timestampToLockDate(endWithdrawTimestamp)
+                : currentBlockTsToLockDate;
+        } else {
+            endWithdrawTimestamp = currentBlockTsToLockDate;
+        }
+
+        if (stopBlock != 0 && endWithdrawTimestamp > stopRewardsTimestamp) {
+            // add 2 weeks because the right boundary is not included in processing cycle below
+            endWithdrawTimestamp = stopRewardsTimestamp.add(TWO_WEEKS);
+        }
```
ororopickpocket: We should discuss this part here. The old code doesn't do it. Does this mean the old code does currently not withdraw all claimable rewards?
If that is the case, you could also use i <= endWithdrawTimestamp in the loop below and save some lines of code - if you make the adjustment I recommended in _comput**e**WeightedStakeForDate.

contracts/governance/StakingRewards/StakingRewardsOs.sol
```diff
@@ -0,0 +1,256 @@
...
+
+    /**
+     * @notice Internal function to calculate weighted stake
+     * @dev Users will receive rewards uo till the stop block
+     * @param _staker Staker address
+     * @param _block Last finalised block
+     * @param _date The date to compute prior weighted stakes
+     * @return The weighted stake
+     * */
+    function _computWeightedStakeForDate(
```
ororopickpocket: typo: _comput**e**WeightedStakeForDate

contracts/governance/StakingRewards/StakingRewardsOsStorage.sol
```diff
@@ -0,0 +1,104 @@
...
+ * @notice Just the storage part of staking rewards contract, no functions,
+ * only constant, variables and required structures (mappings).
+ * Used by StackingRewardsProxy.
+ *
+ * What is SOV staking rewards ?
+ * The purpose of the SOV staking rewards program is to reward,
+ * "marginal stakers" (ie, stakers by choice, not currently vesting) with liquid SOV
+ * at the beginning of each new staking interval.
+ * */
+contract StakingRewardsOsStorage is Ownable {
```
ororopickpocket, line 18: cosnider my comment regarding the ABI changes.

contracts/token/OsSOV.sol
```diff
@@ -0,0 +1,173 @@
...
+ * @notice This contract accounts for all holders' balances.
+ *
+ * @dev This contract represents a token with dynamic supply.
+ *   The owner of the token contract can mint/burn tokens to/from any account
+ *   based upon previous governance voting and approval.
+ *
+ */
+contract OsSOV is ERC20Capped, AccessControl, Ownable, Initializable {
+    string private constant _NAME = "BitcoinOS Sovryn Transition Token";
+    string private constant _SYMBOL = "osSOV";
```
ororopickpocket, line 21: I know it's normal to use constants here, but you might want to consider using variables given that yago wanted to change token names and symbols in the past already.

ororopickpocket (commented):
I recommend a few small changes.

contracts/governance/StakingRewards/StakingRewardsOs.sol
```diff
@@ -0,0 +1,256 @@
...
+    /**
+     * @notice Replacement of constructor by initialize function for Upgradable Contracts
+     * This function will be called only once by the owner.
+     * @param _osSOV osSOV token address
+     * @param _staking StakingProxy address should be passed
+     * */
+    function initialize(
+        address _osSOV,
+        IStaking _staking,
+        uint256 _averageBlockTime
```
cwsnt, line 38: missing natSpec

contracts/governance/StakingRewards/StakingRewardsOs.sol
tjcloa: RewardWithdrawn - reverted
`claimReward` reverted to `collectReward`  
as to the the public variables - i already send ABI to the CoL, they said they adapted it.

contracts/governance/StakingRewards/StakingRewardsOs.sol
```diff
@@ -0,0 +1,256 @@
...
+        require(withdrawTimestamp > 0 && amount > 0, "No valid reward");
+        stakerNextWithdrawTimestamp[msg.sender] = withdrawTimestamp;
+        _payReward(msg.sender, amount);
+    }
+
+    /**
+     * @notice Changes average block time - based on blockchain
+     * @dev If average block time significantly changes, we can update it here and use for block number calculation
+     */
+    function setAverageBlockTime(uint256 _averageBlockTime) external onlyOwner {
```
cwsnt, line 91: missing natspec

contracts/token/OsSOV.sol
tjcloa: i was considering it, just didn't want to break the YAGNI desing principle that you like )  
this is an itermediary temporary token and the contract is upgradable, so, we can just replace it as needed any time

contracts/governance/StakingRewards/StakingRewardsOs.sol
tjcloa: good catch, thanks
but this icheck would be done on every call in a cycle which seems to be better to be moved to the calling function
so, just moved this validation from this function

contracts/governance/StakingRewards/StakingRewardsOs.sol
tjcloa: fixed

contracts/governance/StakingRewards/StakingRewardsOs.sol
```diff
@@ -0,0 +1,256 @@
...
+        osSOV.mint(_staker, _amount);
+        emit RewardPaid(_staker, _amount);
+    }
+
+    /**
+     * @notice Determine the current Block Number
+     * @dev This is segregated from the _getPriorUserStakeByDate function to better test
+     * advancing blocks functionality using Mock Contracts
+     * */
+    function _getCurrentBlockNumber() internal view returns (uint256) {
```
cwsnt, line 160: missing return natspec

contracts/governance/StakingRewards/StakingRewardsOs.sol
```diff
@@ -0,0 +1,256 @@
...
+        return block.number;
+    }
+
+    /**
+     * @notice Internal function to calculate and set block
+     * */
+    function _setBlock(uint256 _checkpointTime) internal {
```
cwsnt, line 167: missing param natspec

contracts/governance/StakingRewards/StakingRewardsOs.sol
```diff
@@ -0,0 +1,256 @@
...
+        return _getStakerCurrentReward(msg.sender, _considerMaxDuration, _startTime);
+    }
+
+    /**
+     * @notice Internal function to calculate staker's current reward
+     * @dev Normally the start time is 0. If this function returns a valid withdraw timestamp
+     * and zero amount - that means there were no valid rewards for that period. So the new period must start
+     * from the end of the last interval or till the time no rewards are accumulated i.e. _startTime
+     */
+    function _getStakerCurrentReward(
```
cwsnt, line 216: missing natspec

contracts/token/OsSOV.sol
```diff
@@ -0,0 +1,173 @@
...
+     */
+    function name() public pure override returns (string memory) {
+        return _NAME;
+    }
+
+    /**
+     * @dev Returns the symbol of the token, usually a shorter version of the
+     * name.
+     */
+    function symbol() public pure override returns (string memory) {
```
cwsnt, line 153: missing return natspec

contracts/token/OsSOV.sol
```diff
@@ -0,0 +1,173 @@
...
+     * @dev Token is non-receivable
+     */
+    receive() external payable {
+        revert NonReceivable();
+    }
+
+    /**
+     * @dev Returns the name of the token.
+     */
+    function name() public pure override returns (string memory) {
```
cwsnt, line 145: missing return natspec

contracts/token/OsSOV.sol
```diff
@@ -0,0 +1,173 @@
...
+     *
+     * Tokens usually opt for a value of 18, imitating the relationship between
+     * Ether and Wei. This is the default value returned by this function, unless
+     * it's overridden.
+     *
+     * NOTE: This information is only used for _display_ purposes: it in
+     * no way affects any of the arithmetic of the contract, including
+     * {IERC20-balanceOf} and {IERC20-transfer}.
+     */
+    function decimals() public pure override returns (uint8) {
```
cwsnt, line 170: missing return natspec

contracts/governance/StakingRewards/StakingRewardsOs.sol
tjcloa: i considered both comments, but decided to keep stop block check here and removed from _computeWeightedStakeForDate - separation of concerns. also, keeping `i` < rewardsInterval.toTimestamp solves potential issues with processing current not finalized block - staking/unstaking, order of txs, i think that was the reason for this solution initially. 
besides, changing this condition breaks all the tests and the tests are rather integration than unit, not comments, many unclear numbers - would take a lot of time to fix all that.

hardhat.config.js
```diff
@@ -13,6 +13,7 @@ require("hardhat-abi-exporter");
 require("hardhat-deploy");
 require("@nomicfoundation/hardhat-chai-matchers");
 require("@nomicfoundation/hardhat-foundry");
+//require("@nomiclabs/hardhat-solhint");
```
cwsnt: can remove this

contracts/governance/StakingRewards/StakingRewardsOs.sol
```diff
@@ -0,0 +1,256 @@
...
+     */
+    function setAverageBlockTime(uint256 _averageBlockTime) external onlyOwner {
+        averageBlockTime = _averageBlockTime;
+    }
+
+    /**
+     * @notice This function computes the last staking checkpoint and calculates the corresponding
+     * block number using the average block time which is then added to the mapping `checkpointBlockNumber`.
+     */
+    function setBlock() external {
```
cwsnt, line 99: Since this function can be called by anyone, I am assuming this function needs to be called periodically?

contracts/governance/StakingRewards/StakingRewardsOs.sol
```diff
@@ -0,0 +1,256 @@
...
+     * @param _date The date to compute prior weighted stakes
+     * @return The weighted stake
+     * */
+    function _computWeightedStakeForDate(
+        address _staker,
+        uint256 _block,
+        uint256 _date
+    ) internal view returns (uint256 weightedStake) {
+        weightedStake = staking.getPriorWeightedStake(_staker, _block, _date);
+        if (stopBlock > 0 && stopBlock < _block) {
```
cwsnt: probably can do this:
```
if(stopBlock == 0 || stopBlock > _block) return weightedStake;

uint256 previousWeightedStake = staking.getPriorWeightedStake(
              _staker,
              stopBlock,
              _date
          );
if (previousWeightedStake < weightedStake) {
    weightedStake = previousWeightedStake;
}

**Review by cwsnt — CHANGES_REQUESTED** [review]

File: contracts/governance/StakingRewards/StakingRewardsOs.sol (hunk 10 above)
Comment (tjcloa) [reply contracts/governance/StakingRewards/StakingRewardsOs.sol:160]: natspec is only required for external and public functions because it is a part of the code documentation and used by docs generators

File: contracts/governance/StakingRewards/StakingRewardsOs.sol (hunk 11 above)
Comment (tjcloa) [reply contracts/governance/StakingRewards/StakingRewardsOs.sol:167]: natspec is only required for external and public functions because it is a part of the code documentation and used by docs generators

File: contracts/governance/StakingRewards/StakingRewardsOs.sol (hunk 8 above)
Comment (tjcloa) [reply contracts/governance/StakingRewards/StakingRewardsOs.sol:38]: fixed

File: contracts/governance/StakingRewards/StakingRewardsOs.sol (hunk 17 above)
Comment (tjcloa) [reply contracts/governance/StakingRewards/StakingRewardsOs.sol:99]: yes, by a cron job - also included in this PR

File: contracts/governance/StakingRewards/StakingRewardsOs.sol (hunk 18 above)
Comment (tjcloa) [reply contracts/governance/StakingRewards/StakingRewardsOs.sol]: stop block would be checked on every iteration - more effective to have this check in the calling function, besides - separation of concerns. refactored.

File: hardhat.config.js (hunk 16 above)
Comment (tjcloa) [reply hardhat.config.js]: removed

File: contracts/governance/StakingRewards/StakingRewardsOs.sol (hunk 3 above)
Comment (tjcloa) [reply contracts/governance/StakingRewards/StakingRewardsOs.sol:190]: added `getArbitraryStakerCurrentReward`

File: contracts/governance/StakingRewards/StakingRewardsOs.sol (hunk 4 above)
Comment (tjcloa) [reply contracts/governance/StakingRewards/StakingRewardsOs.sol]: should not be an issue with the old code because this is ony important when the program actually stops.  
but i will create a ticket to check this assumption.

File: contracts/governance/StakingRewards/StakingRewardsOs.sol (hunk 9 above)
Comment (tjcloa) [reply contracts/governance/StakingRewards/StakingRewardsOs.sol:91]: added param description

File: contracts/governance/StakingRewards/StakingRewardsOs.sol (hunk 12 above)
Comment (tjcloa) [reply contracts/governance/StakingRewards/StakingRewardsOs.sol:216]: natspec required for public functions

File: contracts/token/OsSOV.sol (hunk 15 above)
Comment (tjcloa) [reply contracts/token/OsSOV.sol:170]: it has inherited natspec

File: contracts/token/OsSOV.sol (hunk 14 above)
Comment (tjcloa) [reply contracts/token/OsSOV.sol:145]: added

**Review by tjcloa — COMMENTED** [review]
comments addressed

**Review by ororopickpocket — APPROVED** [review]

File: contracts/governance/StakingRewards/StakingRewardsOs.sol [hunk 19]
```diff
@@ -0,0 +1,284 @@
...
+     */
+    function getArbitraryStakerCurrentReward(
+        bool _considerMaxDuration,
+        uint256 _startTime,
+        address _staker
+    ) external view returns (uint256 nextWithdrawTimestamp, uint256 amount) {
+        return _getStakerCurrentReward(_staker, _considerMaxDuration, _startTime);
+    }
+
+    struct RewardsInterval {
```
cwsnt: better to move the declaration to the top?

cwsnt (approved):
approved with optional comment

contracts/governance/StakingRewards/StakingRewardsOs.sol
tjcloa: fixed

contracts/governance/StakingRewards/StakingRewardsOsStorage.sol
tjcloa: considered

contracts/token/OsSOV.sol
tjcloa: it is there - description of what the function returns

contracts/governance/StakingRewards/StakingRewardsOs.sol
```diff
@@ -73,13 +73,13 @@ contract StakingRewardsOs is StakingRewardsOsStorage, Initializable {
...
         );
-        (uint256 withdrawTimestamp, uint256 amount) = _getStakerCurrentReward(
+        (uint256 nextWithdrawTimestamp, uint256 amount) = _getStakerCurrentReward(
             msg.sender,
             true,
             _startTime
         );
-        require(withdrawTimestamp > 0 && amount > 0, "No valid reward");
-        stakerNextWithdrawTimestamp[msg.sender] = withdrawTimestamp;
+        require(nextWithdrawTimestamp > 0 && amount > 0, "No valid reward");
```
cwsnt, line 81: Can u please cover this edge in the unit test?

PR #543, opened 2024-03-27
Please consider this change to save gas

jjmr007:
Please consider this change to save gas


tjcloa:
the change was applied but modified slightly
thnx

PR #556, opened 2025-07-14
SOV-3587: convert isovryn to interface

cwsnt:


tjcloa:
after careful review, the decision is to close the PR because having ISovryn as a smart contract is obviously for consistency reasons - inheriting events, functions and structs from modules. 
the solution per se is fine, but breaks consistency. 
copypasted events, functions and structs - we’d have to keep it manually updated. 
leads to a lot of unnecessary changes including dependent smart contract.

PR #574, opened 2026-03-03
Pr/feesharingcollector v2 SIP-0088

bribri-wci:
Revert FeeSharingCollector to pre-5429b56 version. Deployed FeeSharingCollector Proxy and Implementations here: 0xd16457C2495406783dECa731f90f6284D09eaFb1 and 0x5D539131c0eb5237753beB7a4d60A3Bb6ecdD96f.


bribri-wci:
@tjcloa Please review, thank you!

tjcloa (changes requested):
While the intent of this PR is to migrate to a new FeeSharingCollector proxy, it currently mixes two incompatible approaches: a pure rollback model (single active collector) and a real migration model (new proxy deployment).   
   
The script deploys a new proxy and implementation (1021-deploy-FeeSharingCollectorV2.js:19), while the canonical deployment still points to the old proxy (FeeSharingCollector.json:2). At the same time, the canonical deployment is rolled back to a previous version. This breaks consistency between the currently active FeeSharingCollector deployment and the deployment metadata.   
   
Effectively, the PR introduces two FeeSharingCollector contracts:    
    
The current deployment (used by scripts and the Sovryn frontend), where user fees remain locked.
A rollback-version deployment that is intended to be switched into protocol usage, but is not supported by the Sovryn app or the scheduled action scripts.  
As a result, this design requires coordinated frontend updates and additional testing so users can claim fees from both contracts while preventing cumulative side effects. Without that, many users are likely to lose access to fees in the orphaned FeeSharingCollector, because withdrawal from that contract may remain economically unprofitable.  

If these changes are not introduced, the frontend will remain  connected to the orphaned FeeSharingCollector, while the newly deployed contracts will require either a separate UI flow or manual direct contract interaction to withdraw fees.   
   
Final conclusion: the solution proposed in this PR is strongly not recommended for acceptance in its current form, because it breaks consistency and introduces significant, unnecessary complexity for Sovryn app integration and long-term maintenance.

